PR #39, opened 2024-10-16
Feat/misc/app theme: Add app color theme and logo

lazarinibruno:
## Description

- This PR focuses on the **theme and logo of the app**. Initially, I wanted to add a complete overhaul of the color theme. However, due to the lack of time before the first milestone, we **decided to stick with the color theme that we already have**. I chose the app logo background color to match the primary color of the color theme.
- I added the new color theme, but it is not in use for the moment. In the future, we hope to refine it and actually use it.
- While researching how to create a theme for an app, I came across many style and theme guidelines that I decided to start summarizing [style guideline](https://github.com/PeriodPals/periodpals/wiki/App-Style-Guide).
## Changes

- Add the color theme that is in development (but not in use)
- Add the logo of the app
- Change the font used to the **system default** to ensure consistency and simplicity.

### Disable dynamic color theme

For those who don't know, newer version of android support **dynamic color themes**. These are themes that automatically match the theme of the user's phone. For example:

image

This is cool as it allows users to control how their app looks and feels. However, I think is best if we stick to a fixed color scheme to create an app identity. But we can discuss this in the future.

## Files Added/Modified/Removed

- `app/src/main/java/com/android/periodpals/ui/theme/Color.kt`
- `app/src/main/java/com/android/periodpals/ui/theme/Theme.kt`
- `app/src/main/java/com/android/periodpals/ui/theme/Type.kt`
- `app/src/main/res/<android studio modified some files here for the new logo>
`
## Dependencies Added

None.

## Testing

No testing to be done.

## Screenshots

drawing

drawing

### DIFF
--- a/app/src/main/java/com/android/periodpals/ui/theme/Color.kt
+++ b/app/src/main/java/com/android/periodpals/ui/theme/Color.kt
@@ -47,7 +47,6 @@ val surfaceContainerLight = Color(0xFFEDEDF4)
 val surfaceContainerHighLight = Color(0xFFE7E8EE)
 val surfaceContainerHighestLight = Color(0xFFE2E2E9)
 
-
 /* Dark theme, still in development and thus not in use */
 val primaryDark = Color(0xFFAAC7FF)
 val onPrimaryDark = Color(0xFF0A305F)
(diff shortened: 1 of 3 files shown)


taghizadlaura (commented):
### Important 

* Code quality : LGTM
* Functionality : LGTM
* Performance : LGTM
* Security : LGTM

### Testing

NTR
We'll have to do that ourselves later, as it mainly concerns design/UI.

### Documentation 

LGTM

### Suggestions

As I mentioned in the comments of the review, since we will implement the dynamic colour scheme and typography later (in another sprint), should we keep the code commented out in these sections? Wouldn't it make the code cleaner to remove it for now?

PR #40, opened 2024-10-16
Revert "Feat/misc/app theme: Add app color theme and logo"

charliemangano:
### Description
This PR reverts the unintentional rebasing of unreviewed #39 onto main (sorry everyone).

### Changes
Reverted the changes made in PR #35.

#### Added
Resource files (default android app logo).
#### Modified
 - `app/src/main/java/com/android/periodpals/ui/theme/Type.kt`
 - `app/src/main/java/com/android/periodpals/ui/theme/Color.kt`
 - `app/src/main/java/com/android/periodpals/ui/theme/Theme.kt`
#### Removed
Resource files (logo designed by @lazarinibruno and binaries).

#### Dependencies Added
None

#### Testing
Not applicable

#### Screenshots
Not applicable

PR #41, opened 2024-10-16
Feat/misc/app theme: Add app color theme and logo

charliemangano:
**This is a copy of the PR #39 by @lazarinibruno**. It was unintentionally merged without having been reviewed and therefore reverted (#40).

## Description

- This PR focuses on the **theme and logo of the app**. Initially, I wanted to add a complete overhaul of the color theme. However, due to the lack of time before the first milestone, we **decided to stick with the color theme that we already have**. I chose the app logo background color to match the primary color of the color theme.
- I added the new color theme, but it is not in use for the moment. In the future, we hope to refine it and actually use it.
- While researching how to create a theme for an app, I came across many style and theme guidelines that I decided to start summarizing [style guideline](https://github.com/PeriodPals/periodpals/wiki/App-Style-Guide).
## Changes

- Add the color theme that is in development (but not in use)
- Add the logo of the app
- Change the font used to the **system default** to ensure consistency and simplicity.

### Disable dynamic color theme

For those who don't know, newer version of android support **dynamic color themes**. These are themes that automatically match the theme of the user's phone. For example:

image

This is cool as it allows users to control how their app looks and feels. However, I think is best if we stick to a fixed color scheme to create an app identity. But we can discuss this in the future.

## Files Added/Modified/Removed

- `app/src/main/java/com/android/periodpals/ui/theme/Color.kt`
- `app/src/main/java/com/android/periodpals/ui/theme/Theme.kt`
- `app/src/main/java/com/android/periodpals/ui/theme/Type.kt`
- `app/src/main/res/<android studio modified some files here for the new logo>
`
## Dependencies Added

None.

## Testing

No testing to be done.

## Screenshots

drawing

drawing


app/src/main/java/com/android/periodpals/ui/theme/Type.kt
```diff
@@ -1,13 +1,14 @@
...
-import androidx.compose.ui.text.font.FontWeight
-import androidx.compose.ui.unit.sp
 
 // Set of Material typography styles to start with
-val Typography =
+
+// For the moment, use the device default typography
+val Typography = Typography()
+
+// The app typography is still in development
```
taghizadlaura, line 10: LGTM. It would be nice to have app-specific typography in the future to avoid inconsistencies in text presentation throughout the app!

app/src/main/java/com/android/periodpals/ui/theme/Theme.kt
```diff
@@ -44,13 +119,25 @@ fun PeriodPalsAppTheme(
...
       when {
-        dynamicColor && Build.VERSION.SDK_INT >= Build.VERSION_CODES.S -> {
-          val context = LocalContext.current
-          if (darkTheme) dynamicDarkColorScheme(context) else dynamicLightColorScheme(context)
-        }
         darkTheme -> DarkColorScheme
         else -> LightColorScheme
       }
+
+  // Dynamic color scheme (disabled for the moment)
```
taghizadlaura, line 126: It would be nice to integrate the dynamic colour scheme as well, but since it's planned for later, do we need to keep all this code in comments (same for the typography below)? Wouldn't it be better to start this in another task in a future sprint?

app/src/main/java/com/android/periodpals/ui/theme/Color.kt
```diff
@@ -9,3 +9,77 @@ val Pink80 = Color(0xFFEFB8C8)
 val Purple40 = Color(0xFF6650a4)
 val PurpleGrey40 = Color(0xFF625b71)
 val Pink40 = Color(0xFF7D5260)
+
+/* Light theme, still in development and thus not in use */
```
taghizadlaura, line 13: Nice colour palette !

taghizadlaura (commented):
### Important  

* Code quality : LGTM 
* Functionality : LGTM 
* Performance : LGTM 
* Security : LGTM 

### Testing 

NTR 
We'll have to do that ourselves later, as it mainly concerns design/UI. 

### Documentation  

LGTM 

### Suggestions 

As I mentioned in the comments of the review, since we will implement the dynamic colour scheme and typography later (in another sprint), should we keep the code commented out in these sections? Wouldn't it make the code cleaner to remove it for now?

app/src/main/java/com/android/periodpals/ui/theme/Type.kt
lazarinibruno: I totally agree! I started looking at some possible fonts. Maybe I can put some photos in the styling guide so that the team can start looking at them. Also, if you have suggestions for the fonts, put them in the guide!

app/src/main/java/com/android/periodpals/ui/theme/Theme.kt
lazarinibruno: Yes, I was hesitant about commenting out instead of simply removing it. However, I thought it would be practical to leave the code commented. Like this, whenever we want to see how the new theme looks, its just matter of uncommenting it.

For the font I just commented it because I did not know what it was used for, so I did not want to delete it. But in retrospective, I should have asked what it is for. I will update the PR with the question.

lazarinibruno: I asked around the team members and no one had added the commented code in `Type.kt`, so I will delete it.

taghizadlaura (approved):
LGTM now :)

agonzalez-r (approved):
## Summary
Great work on the app aesthetics, I think it's good to implement this early on the project for better adaptability of the work in the following sprints.  

## Important 
- Code quality:  LGTM
- Functionality: LGTM 
- Security: LGTM
- Performance: LGTM

## Testing
Not yet implemented. 

## Documentation
Very well structured and explained Style Guide on the wiki. This will facilitate the implementation for all the members on the team. 

## Suggestions
I suggest not implementing the Dynamic Colour Scheme at all, I don't think it's aligned with our app's goals to have that UI personalisation, since the app it's about community and sharing it's best to have a common theme between users.

lazarinibruno:
> I suggest not implementing the Dynamic Colour Scheme at all, I don't think it's aligned with our app's goals to have that UI personalisation, since the app it's about community and sharing it's best to have a common theme between users.

@agonzalez-r I agree with you. Upon discussing it with the team, the other members agree that we wont use it. I'll delete the commented text then.

PR #58, opened 2024-10-18
fix: resolve navigation and sign-in and sign-up flows

francelu:
Assure that the M1 flows work.


taghizadlaura (approved):
LGTM. Sign up with wrong password is not perfect.

charliemangano (approved):
looks good to me. we will come back to it to clean up at a late date

Issue #64, opened 2024-10-27
Fix UI tests for M1: Alert


charliemangano:
See #63. This task is specifically for the Alert screen.

### Done
* [x] UI tests are complete
* [x] Line coverage is > 80%

Issue #63, opened 2024-10-27
Fix UI tests for M1


charliemangano:
Fix and clean up UI tests for the files submitted for M1. Add missing tests.

### Done
* [x] UI tests complete for all the screens:
  * [x] Alert (#64, #68)
  * [x] AlertList (#69, #70)
  * [x] SignIn (#71, #73)
  * [x] SignUp (#76, #96)
  * [x] CreateProfile (#74)
  * [x] Profile (#83)
* [x] UI tests complete for all the navigation components:
  * [x] BottomNavigationMenu (#97, #105)
  * [x] TopAppBar (#90, #91)
* [x] Line coverage is > 80%

Map : We decided to complete the tests when the task "Change Map to Maplib" will be done
MainActivity (#86) : We decided to do not fix it for this week since the existing tests were about the Location which will be moved in another file

agonzalez-r:
First step: 

Get Mockito to work for Android Tests in order to mock navigationActions

francelu:
Hey, in #66 I'm checking the onClick actions. Don't bother too much about this part since I'll be writing tests for that specific part too. Mainly `CreateProfileScreen` and `CreateAlertScreen`.

PR #68, opened 2024-10-27
Fix/alert/UI tests: Fix and complete UI tests for AlertScreen

charliemangano:
# Fix and complete UI tests for AlertScreen
## Description
This PR introduces new tests for the Alert screen and fixes the old ones. It also fixes the problems pointed out by said new tests. It closes issue #64 and fixes part of #63 .
## Changes
Added dependencies to make mockito work in androidTests, and properly mocked `NavigationActions`. Added verification of correct navigation in the old tests and created new tests to check that nothing happens when trying to submit an invalid Alert.  
Invalid alerts are ones which don't have a period product, urgency level and location selected.
#### Added
None
#### Modified
- `app/build.gradle.kts` and `gradle/libs.versions.toml`: add the missing dependency
- `app/src/androidTest/java/com/android/periodpals/ui/alert/AlertScreenTest.kt`: modified and added tests
- `app/src/main/java/com/android/periodpals/ui/alert/AlertScreen.kt`: fixed problems pointed out by new tests
#### Removed
None
## Dependencies Added
- [`com.linkedin.dexmaker:dexmaker-mockito-inline`](https://stackoverflow.com/a/67337292) to make mockito work with androidTests. Had to remove `libs.mockito.android` for androidTests as a consequence
## Testing
All UI tests for `AlertScreen` pass. They should be complete enough.

### DIFF
--- a/app/src/androidTest/java/com/android/periodpals/ui/alert/AlertScreenTest.kt
+++ b/app/src/androidTest/java/com/android/periodpals/ui/alert/AlertScreenTest.kt
@@ -50,9 +50,9 @@ class AlertScreenTest {
     composeTestRule.onNodeWithTag("topBar").assertIsDisplayed()
     composeTestRule.onNodeWithTag("goBackButton").assertIsNotDisplayed()
     composeTestRule
-      .onNodeWithTag("alertSubmit")
-      .assertIsDisplayed()
-      .assertTextEquals("Ask for Help")
+        .onNodeWithTag("alertSubmit")
+        .assertIsDisplayed()
+        .assertTextEquals("Ask for Help")
   }
 
   @Test
--- a/app/src/androidTest/java/com/android/periodpals/ui/alert/AlertScreenTest.kt
+++ b/app/src/androidTest/java/com/android/periodpals/ui/alert/AlertScreenTest.kt
@@ -1,26 +1,39 @@
 package com.android.periodpals.ui.alert
 
-import androidx.compose.material3.MaterialTheme
 import androidx.compose.ui.test.assertIsDisplayed
 import androidx.compose.ui.test.assertTextEquals
 import androidx.compose.ui.test.junit4.createComposeRule
 import androidx.compose.ui.test.onNodeWithTag
+import androidx.compose.ui.test.onNodeWithText
 import androidx.compose.ui.test.performClick
 import androidx.compose.ui.test.performTextInput
-import androidx.navigation.compose.rememberNavController
+import androidx.test.ext.junit.runners.AndroidJUnit4
 import com.android.periodpals.ui.navigation.NavigationActions
+import com.android.periodpals.ui.navigation.Route
+import com.android.periodpals.ui.navigation.Screen
+import org.junit.Before
 import org.junit.Rule
 import org.junit.Test
+import org.junit.runner.RunWith
+import org.mockito.Mockito.mock
+import org.mockito.Mockito.`when`
+import org.mockito.kotlin.verify
 
+@RunWith(AndroidJUnit4::class)
 class AlertScreenTest {
 
+  private lateinit var navigationActions: NavigationActions
   @get:Rule val composeTestRule = createComposeRule()
 
+  @Before
+  fun setUp() {
+    navigationActions = mock(NavigationActions::class.java)
+    `when`(navigationActions.currentRoute()).thenReturn(Route.ALERT)
+  }
+
   @Test
-  fun displayAllComponents() {
-    composeTestRule.setContent {
-      MaterialTheme { AlertScreen(NavigationActions(rememberNavController())) }
-    }
+  fun allComponentsAreDisplayed() {
+    composeTestRule.setContent { AlertScreen(navigationActions) }
 
     composeTestRule.onNodeWithTag("alertInstruction").assertIsDisplayed()
     composeTestRule.onNodeWithTag("alertProduct").assertIsDisplayed()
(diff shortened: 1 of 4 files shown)


app/src/androidTest/java/com/android/periodpals/ui/alert/AlertScreenTest.kt
```diff
@@ -34,22 +50,74 @@ class AlertScreenTest {
...
+    composeTestRule.waitForIdle()
 
     composeTestRule.onNodeWithTag("alertLocation").performTextInput("Rolex")
     composeTestRule.onNodeWithTag("alertMessage").performTextInput("I need help finding a tampon")
 
-    // Cannot test navigation actions
-    //    composeTestRule.onNodeWithTag("alertSubmit").performClick()
+    composeTestRule.onNodeWithTag("alertSubmit").performClick()
+    verify(navigationActions, never()).navigateTo(any<TopLevelDestination>())
+    verify(navigationActions, never()).navigateTo(any<String>())
```
agonzalez-r, line 83: This line seems redundant since the previous one already checks that no navigation action was called when clicked on the submit button with invalid inputs

app/src/main/java/com/android/periodpals/ui/alert/AlertScreen.kt
```diff
@@ -35,74 +37,102 @@ import com.android.periodpals.ui.navigation.TopAppBar
 
 @Composable
 fun AlertScreen(navigationActions: NavigationActions) {
+  val context = LocalContext.current
   var location by remember { mutableStateOf("") }
   var message by remember { mutableStateOf("") }
+  val (productIsSelected, setProductIsSelected) = remember { mutableStateOf(false) }
+  val (urgencyIsSelected, setUrgencyIsSelected) = remember { mutableStateOf(false) }
```
francelu, line 44: I don't understand this double instantiation required for respectively product and urgency.

francelu (changes requested):
LGTM.  
Look at the code comments. You need to check the top bar is displayed.

app/src/main/java/com/android/periodpals/ui/alert/AlertScreen.kt
agonzalez-r: it serves in order to check that one option for each field was selected, to check if the alert is valid or not (if it's not valid, not possible to submit and toast is displayed)

agonzalez-r (approved):
## Summary
Great work! This PR effectively fixes the dependencies issues in order for Mockito to work on Android Tests, and thoroughly tests `AlertScreens`. 

## Important 
#### Code Quality
- `AlertScreenTest`: check comment
- (`build.gradle.kts`: this whole file should be cleaned (lots of commented / repeated code), but not the scope of this PR)

## Suggestions
- Future improvements: when data class `Alert` is implemented, remove the two variables `productIsSelected` and `urgencyIsSelected` and check if the fields are non-empty instead

app/src/main/java/com/android/periodpals/ui/alert/AlertScreen.kt
charliemangano: See the [three ways to declare a `MutableState` in a composable](https://developer.android.com/develop/ui/compose/state#state-in-composables). In short, it's equivalent to the other ways, but uses `val`s (which I personally prefer and was needed here).   

Instead of having this (v1): 
```kotlin
var productIsSelectedv1 by remember {mutableStateOf(false)} // init var to false
productIsSelectedv1 = true // change its value in `ExposedDropdownMenuSample`
``` 
you get this (v2):
```kotlin
val (productIsSelectedv2, setProductIsSelected) = remember {mutableStateOf(false)} // init productIsSelectedv2 to false
setProductIsSelected(true) // change its value in `ExposedDropdownMenuSample`
```
where `productIsSelectedv2` is a `Boolean` and `setProductIsSelected` is a `(Boolean) -> Unit`.

I had to use this here because we want to change its value in `ExposedDropdownMenuSample`, so we have to pass it as an argument. However, Kotlin passes arguments by-value, so passing `productIsSelectedv1` (the `var`) to the function doesn't make it mutable inside the function (trying `isSelectedv1 = true` in `ExposedDropdownMenuSample` throws an exception because it has essentially become a `val`)

Does this answer the question?

charliemangano:
Thanks for your review @agonzalez-r ! I'll answer to some things that you pointed out quickly:
> * (`build.gradle.kts`: this whole file should be cleaned (lots of commented / repeated code), but not the scope of this PR)

Can you create an issue about this? We'll get to it when we have time.
  

> * Future improvements: when data class `Alert` is implemented, remove the two variables `productIsSelected` and `urgencyIsSelected` and check if the fields are non-empty instead

I don't believe this is relevant as checking basic properties about field content should be done at the View level and not by the ViewModel or Model because of performance issues (see this [Ed post](https://edstem.org/eu/courses/1567/discussion/129490))

app/src/main/java/com/android/periodpals/ui/alert/AlertScreen.kt
francelu: Good, thanks that is very clear !

app/src/androidTest/java/com/android/periodpals/ui/alert/AlertScreenTest.kt
charliemangano: As there are two `navigateTo` functions (one with a `TopLevelDestination` as argument and another with a `String`), I need to check both overloads.
I searched for a way to have it as a one-liner by not specifying the type at all or having something like `<TopLevelDestination || String>` but it didn't work (respectively because: it throws an exception, and: I couldn't find the proper syntax)

app/src/androidTest/java/com/android/periodpals/ui/alert/AlertScreenTest.kt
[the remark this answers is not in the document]
charliemangano: Done! Can you re-review?

francelu (approved):
LGTM ! :)

PR #85, opened 2024-10-28
Feat/geolocation/location-service : Add Location Services Interface and Implementation

lazarinibruno:
# Add Location Services Interface

## Description
This PR introduces an interface `LocationService` and its implementation `LocationServiceImpl`. For the moment, they handle asking for location access to the user. But in the future, it will also handle the user location.

The motivation for this PR was that the location access dialog when launching the app made the E2E test fail. So me and @charliemangano thought it would be a good idea to finally have a proper location service.

I used this website for reference: [Request location permissions](https://developer.android.com/develop/sensors-and-location/location/permissions).

## Changes
- Manifest
    - Updated comments to make more transparent what each permission request is doing.
    - Added the permission to access the location in  the background.

#### Added
- `LocationService.kt` Interface for the location services
- `LocationServiceImpl.kt` Implementation of the interface

#### Modified
- `AndroidManifest.xml`
- `build.gradle.kts`

## Dependencies Added
- `libs.androidx.activity.ktx`: Added to facilitate permission management from the activity.
- `libs.androidx.fragment.ktx.v184`: Required to use fragments effectively while managing permissions.

## Implementation Details

- The constructor of the `LocationServiceImpl` takes as an argument a `ComponentActivity`. This is because:
    - Permission requests in Android must be initiated from an Activity component, because they need to show a system dialog to the user.
    - In other words, we can think of an Activity as a screen. The system dialog asking for the permission needs to know in what screen it should be displayed.
    - By passing the Activity to the service, we tell it "use this window to show the permission dialog".

So, in the MainActivity, we need to add the following line:

```kotlin
// Initialize the location service
locationService = LocationServiceImpl(this)
```

And every screen using location services must have the location service as argument:

```kotlin
@Composable
screenUsingLocation(locationService: LocationService)
```

## Testing
I finally finished the tests. It took a good while because I was having trouble capturing a callback using Mockito, but it now works :)

The tests check that the location service correctly updates the `StateFlow` that contain the granted location type.
## Test implementation details

I learnt a lot when writing these tests. To make sure I understand and to potentially help others understand, I will explain what the `LocationServiceImplTest.kt` is doing.

### Initializing the mocked classes

First of all, when unit testing a class, everything should be mocked but for that class. Doing so allows us to isolate that class and avoid side-effects of other classes (yes, I just realized this). In this case, everything but the `LocationServiceImpl` is mocked.

The most important part for these tests is capturing the callback that `registerForActivityResult` takes as argument in `LocationServiceImpl`:

```kotlin
private val requestPermissionLauncher: ActivityResultLauncher<Array<String>> =
      activity.registerForActivityResult(ActivityResultContracts.RequestMultiplePermissions()) {
          permissions -> {} // the famous captured callback
```

Mockito has a tool called `Captor` which allows us to capture arguments passed into a function. In this case, we have:

```kotlin
@Captor
private lateinit var permissionCallbackCaptor:
        ArgumentCaptor< ActivityResultCallback<(Map<String, Boolean> >>
```

Capturing the callback allows us to later test scenarios in which the user granted different permission types.

### setup() function

First, we mock the RegisterForActivityResult call:

```kotlin
doReturn(permissionLauncher).`when`(activity).registerForActivityResult(
  any<ActivityResultContracts.RequestMultiplePermissions>(),
  any<ActivityResultCallback<Map<String, Boolean>>>()
)
```

When mocking a function, Mockito needs to know when to apply that behaviour. Here, we tell Mockito to return `permissionLauncher` when `registerForActivityResult` is called with:

- `any<ActivityResultContracts.RequestMultiplePermissions>()` and
  
- `any<ActivityResultCallback<Map<String, Boolean>>>()`
  

Then, the location service is instantiated

```kotlin
locationService = LocationServiceImpl(activity)
```

And immediately after we check that `registerForActivityResult` was called, and capture the callback:

```kotlin
verify(activity).registerForActivityResult(
  any<ActivityResultContracts.RequestMultiplePermissions>(),
  capture(permissionCallbackCaptor)
)
```

### Testing

The interesting tests are those that start with "should update to...". All of them test the behaviour when different permissions are granted. They all start by simulating which permissions were granted by the user:

```kotlin
val permissions = mapOf(
    Manifest.permission.ACCESS_FINE_LOCATION to true,
    Manifest.permission.ACCESS_COARSE_LOCATION to true
)
```

Then, we can "run" the callback we had previously captured:

```kotlin
permissionCallbackCaptor.value.onActivityResult(permissions)
```

The callback takes as argument the permissions sets the `_locationGrantedType` (in `LocationServiceImpl.kt`) to `PRECISE`.

Finally, we check that the callback indeed worked:

```kotlin
assertEquals(LocationAccessType.PRECISE, locationService.locationGrantedType.first())
```


app/src/test/java/com/android/periodpals/services/LocationServiceImplTest.kt
```diff
@@ -0,0 +1,122 @@
...
+    // Then
+    verify(permissionLauncher)
+        .launch(
+            arrayOf(
+                Manifest.permission.ACCESS_FINE_LOCATION,
+                Manifest.permission.ACCESS_COARSE_LOCATION))
+  }
+
+  @Test
+  fun `should update to PRECISE wehn fine location permission is granted`() = runBlocking {
```
agonzalez-r, line 76: Nitpick: typo in the name of the test

app/src/test/java/com/android/periodpals/services/LocationServiceImplTest.kt
```diff
@@ -0,0 +1,122 @@
...
+    // And then we verify that upon creating the location service,
+    // registerForActivityResult was called. We also capture the callback.
+    verify(activity)
+        .registerForActivityResult(
+            any<ActivityResultContracts.RequestMultiplePermissions>(),
+            capture(permissionCallbackCaptor))
+  }
+
+  @Test
+  fun `initial location access type should be NONE`() = runBlocking {
```
agonzalez-r, line 58: Nitpick: the naming style of the tests doesn't match the rest of them, which is not good for consistency... On the other hand this style is also [common practice](https://kotlinlang.org/docs/coding-conventions.html#names-for-test-methods) in Kotlin (and easier to read), so it's up to you to leave it this way

app/src/main/java/com/android/periodpals/MainActivity.kt
```diff
@@ -101,23 +100,23 @@ class MainActivity : ComponentActivity() {
...
-        grantResults.isNotEmpty() &&
-        grantResults[0] == PackageManager.PERMISSION_GRANTED) {
-      // **Permission granted, update state**
-      locationPermissionGranted = true
-    } else {
-      // **Permission denied, notify user**
-      Toast.makeText(this, "Location permission denied.", Toast.LENGTH_SHORT).show()
-    }
-  }
+  //  @Deprecated("Deprecated in Java")
```
agonzalez-r, line 103: Nitpick: commented code should be removed. But there is a task to clean up all files anyways

agonzalez-r (approved):
## Summary
Thank you for your work on this! It works greats. 
(And also thank you for your very detailed PR description!! It made it way easier to understand)

#### Code Quality
Very clean code overall. 
- I left a comment about the naming styles of the tests (good for readability but not consistent)
#### Functionality
- I noticed there is no error handling in the code. I looked at the [documentation that you linked](https://developer.android.com/develop/sensors-and-location/location/permissions#request-location-access-runtime) and they don't mention error handling so what you did is correct. But I would also consider encapsulating the `when{}` in a `try{}` and catch the error...

## Testing
- Tests are correct but I couldn't find the coverage on SonarCloud...
## Suggestions
- I also looked at this section of the documentation [Explain why your app needs the permission](https://developer.android.com/training/permissions/requesting#explain) (in the case they decline) and I think this could be a great improvement.

taghizadlaura (approved):
## Summary
LGTM but please look the suggestion section.

## Important

#### Code Quality
LGTM :
- The code is generally well-structured and organized, with clear imports and logical separation of concerns.
- Enum usage for `LocationAccessType` enhances readability.
- Inline comments provide helpful context, but additional documentation would improve understanding.

#### Functionality
LGTM :
- The implementation correctly manages permission requests and updates the location access state.

#### Security
- Security implications related to handling permissions seems appropriately considered, but error handling for permission requests could enhance robustness.

## Testing
LGTM :
- Tests are comprehensive, using Mockito to mock dependencies effectively.
- Each test is clearly named and covers expected outcomes based on different permission states.
- Ensure to test for edge cases to guarantee stability in various scenarios.

## Documentation
- LGTM

## Suggestions
- Maybe adding tests for edge cases and unexpected outcomes, such as permission denial and user cancellation.
- Maybe Implement error handling for permission request failures (but I don't know how to do it).
- Refactor permission strings into constants for better maintainability, for example : 
```kotlin
object Permissions {
    const val LOCATION_FINE = Manifest.permission.ACCESS_FINE_LOCATION
    const val LOCATION_COARSE = Manifest.permission.ACCESS_COARSE_LOCATION
}
```


## Steps before PR approved
LGTM

taghizadlaura (commented):
I think I don't understand everything you implemented but if the PR corresponds to the issue #95 , when I run the MainActivity on the emulator this doesn't match the requirements :
Location access should be asked for when the user interacts with a part of the app that needs location access. For instance, when the user clicks on the map.

When I ran it, it asked me my Location on the Authentication screen and when I went in the Map it localised me at the default location instead of my real one (I selected the option "Only this time")

app/src/main/java/com/android/periodpals/MainActivity.kt
lazarinibruno: Oh, I commented this code since it had an error and it didn't let me run the unit tests. However, I did not modify any logic in the MainActivity. I think it is best to integrate the LocationService to the app in another PR.

lazarinibruno:
> I think I don't understand everything you implemented but if the PR corresponds to the issue #95 , when I run the MainActivity on the emulator this doesn't match the requirements : Location access should be asked for when the user interacts with a part of the app that needs location access. For instance, when the user clicks on the map.
> 
> When I ran it, it asked me my Location on the Authentication screen and when I went in the Map it localised me at the default location instead of my real one (I selected the option "Only this time")
> 
> Is this the expected behaviour ?

For the moment, I just implemented the location service and wrote tests for it. The code that I commented in the MainActivity was already there (I did not write it). I commented it  because there is an error in it, and it did not let me launch the tests.

app/src/test/java/com/android/periodpals/services/LocationServiceImplTest.kt
lazarinibruno: You are right, but given that the names for the tests are quite long, I preferred readability than consistency with other test names

lazarinibruno:
> * I noticed there is no error handling in the code. I looked at the [documentation that you linked](https://developer.android.com/develop/sensors-and-location/location/permissions#request-location-access-runtime) and they don't mention error handling so what you did is correct. But I would also consider encapsulating the `when{}` in a `try{}` and catch the error (in `LocationServiceImpl.kt`)...

This is a very good idea! I just implemented it ;)

lazarinibruno:
> ## Suggestions
> 
>     * Maybe adding tests for edge cases and unexpected outcomes, such as permission denial and user cancellation.
> 
>     * Maybe Implement error handling for permission request failures (but I don't know how to do it).
> 
>     * Refactor permission strings into constants for better maintainability, for example :

- Permission denial is already covered by the last test. For user cancellation, I think it is equivalent to denial, since by default, the location access is not granted.
- As I said to @agonzalez-r, this is a good idea :) I just implemented it.
- I don't really see what you mean by this.

Issue #82, opened 2024-10-28
Make layout fit to all screen sizes


francelu:
Here is the feedback from the M1:
```markdown
## UI

We checked whether the elements of your app's UI are consistent (fonts, colors, etc.) and whether the layout is responsive (i.e., adapts to different screen sizes and orientations).

It does not seem to fit all screen orientations (you can block it otherwise to portait), and the dark mode makes texts less readable.

For this part, you received 3 points out of a maximum of 5.
```  
---

### **Description:**

The following task description is an after thought from #93.

This task focuses on updating the layout for the following screens to ensure they adapt properly to all screen sizes:
- Authentication:
  - [x] #92
    PR #136
  - [x] #94
    PR #142
- Profile:
  - [x] #169
    PR  #170
  - [x] #172
    PR  #174
  - [x] #175
    PR #176
    PR #177
- Alerts:
  - [x] #147
    PR #153
  - [x] #166
    PR #173
- Navigation:
  - [x] #144
    PR #146

The main objectives are to standardise padding, dimensions, and typography across all screens, improving UI consistency and alignment with design guidelines. This aligns with our sprint goal of enhancing the app's usability and responsiveness across device sizes.

Additionally, a page will be published on the wiki detailing the specific padding and typography guidelines to use in different components, ensuring consistent UI design across the project.

### **Acceptance Criteria:**
- **Functional Requirements:** 
  - Adjust all padding, dimensions, and typography to be responsive based on screen size.
  - Ensure all UI elements are visible and accessible on both small and large screens.
  - Confirm consistency in spacing and font sizes across each screen.
  
- **Quality Standards:** 
  - Follow design standards from Figma and match layout specifications.
  - Ensure code quality through adherence to established coding practices.

- **Testing Requirements:** 
  - Conduct unit and integration testing to confirm that layout adjustments work consistently across screen sizes.
  - Code coverage for UI components should meet or exceed 80%.

### **Preconditions:**  
- Reference Figma design specifications for alignment on dimensions and typography.
- Ensure any required dependencies, such as updated design tokens or theming files, are available.

### **Steps to Completion:**
- [x] #103
- [x] **Step 2:** Adjust padding across all specified screens to ensure consistent appearance on different screen sizes.
  - [x] #158
    PR #165
  - [x] #157
    PR #168
- [x] **Step 3:** Test each screen for responsiveness and verify that elements scale properly.
[...]

PR #93, opened 2024-10-30
Fix/style/layout responsiveness/set up

francelu:
# Set Up Responsive Dimensions and Typography for All Screens

## Description
This PR establishes a foundation for responsive design by setting up scalable dimensions and typography that adapt to different screen sizes. This setup allows screens to adjust automatically, ensuring consistency across devices. Subtask from #82.

## Changes
- Defined responsive dimensions and typography settings.
- Established scalable padding, font sizes, and layout parameters according to screen size for use across all screens.

#### Added
- To be determined.

## Dependencies Added
- `androidx.compose.material3:material3-window-size-class:1.3.0` for window size class handling to support responsive layouts.

## Testing
- Verified that dimensions and typography scale correctly across various screen sizes without issues on existing screens.

## Screenshots (if applicable)
*Include examples of the scalable elements in action if needed.*

### DIFF
--- a/app/src/main/java/com/android/periodpals/ui/authentication/SignIn.kt
+++ b/app/src/main/java/com/android/periodpals/ui/authentication/SignIn.kt
@@ -15,6 +15,7 @@ import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.layout.wrapContentSize
+import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.foundation.text.ClickableText
 import androidx.compose.material3.Button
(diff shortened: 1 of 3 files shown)


charliemangano (changes requested):
## Summary
Nice work on this! I tested it on my own device (pup-up window to resize) and it works seamlessly on most window sizes, with some exceptions.
I noticed you used `LazyColumn`s instead of normal `Column`s, more on this in code comment.

## Important
#### Code Quality
The code is generally unchanged, with modifications made to the padding values and `Column` changes to `LazyColumn`. Out of the scope of this issue but we'll need to clean it up a bit.

I believe padding management should be relegated to the `Theme.kt`, where you put default values for XS to XL-size padding and apply them everywhere. [This 9min video](https://youtu.be/9ElrcTtAxzA) explains it pretty well, but sadly it does not go Jetpack Compose into implementation detail.
This would make it easier to standardize across screens as well, so I think any work you'd put into it now is time gained later on.
#### Functionality
Because the emulator doesn't work on my device, I ran the app directly on my phone and used the pop-up and split-screen view to try and resize it. It works very well on all **full-width sized windows** (where it touches both left and right borders), but when resizing it with small width, the layout is still shaky:
**At ~3/4 of normal width:**
3/4 width layout bug
**At ~1/2 of normal witdh:**
1/2 width layout bug
For reference, my device screen size is a Samsung A32 (size:	6.4 inches, resolution:	1080 x 2400 pixels, ratio: 20:9).
## Testing
All UI tests still pass
## Documentation
Not applicable (not the scope of this issue but will need reworking)
## Suggestions
I believe either setting a minimum width for the app or having dynamically resized text would solve the layout issue with small widths, but I have not done any further research on the subject. 
## Steps before PR approved
* Fix layout with small widths
* Standardize padding using `Theme.kt`

francelu:
@charliemangano  
I've found a great video "The Ultimate guide to supporting all screen sizes in Jetpack Compose using Material 3 design" on [YouTube](https://www.youtube.com/watch?v=DeumyOzKqgI&list=RDVoEsEC2CLgE&index=2).  
However, then I would have to also change all  the typography. We should choose a font too.

charliemangano:
@francelu there is already a default font defined in `Type.kt`. I believe we can use that one until we work on a more precise theme

Issue #84, opened 2024-10-28
Clean up code from M1


charliemangano:
Clean up the code from the first two sprints.

### Done:
* All nomenclature is consistent (between classes and more largely with the Figma)
* Clean code, respects conventions
* Logical file hierarchy
* Documentation
* Across all files of all packages (from M1), including unit and integration tests

The code packages (or if not applicable, single files) are:
* [x] #87, #88 
* [x] #99, #100 
* [x] #101, #102
* [x] #122, #123
* [x] #124, #125 
* [ ] `com.android.periodpals.ui.components`
* [ ] `com.android.periodpals.ui.navigation`
* [ ] `com.android.periodpals.ui.profile`
* [ ] `com.android.periodpals.ui.theme`
* [ ] `com.android.periodpals.ui.timer`
* [ ] `com.android.periodpals.MainActivity.kt`

Code related to the map and the backend management of user profiles will not be done as the correct implementation will be replacing it soon.

charliemangano:
We're focusing on higher priority tasks for now, might come back to this later but don't count on it.

PR #98, opened 2024-10-30
Fix/profile/profile UI tests: Fix and complete tests for ProfileScreen 

agonzalez-r:
# Fix and complete tests for ProfileScreen 

## Description
This PR closes issue #83. It completes the `ProfileScreen` by adding the edit button (on the top bar) that navigates to `EditProfileScreen`.  Tests were also modified to test all functionalities of this screen (mostly the correct displays of component) and added a test for the navigation of the edit button. 

#### Added
- none
#### Modified
- `ProfileScreen`:  modified UI by adding a card when there are no reviews for the profile (matching the Figma) and 
- `ProfileScreenTest`: used proper navigation mock for the tests using `Mockito`, tests all components in the screen

#### Removed
- none

## Tests
- `ProfileScreenTest`: added a test to ensure that the edit button navigates to `EditProfileScreen` when clicked 

## Dependencies Added
- none

## Screenshots 
Screenshot 2024-10-30 at 15 46 30

### DIFF
--- a/app/src/androidTest/java/com/android/periodpals/ui/profile/ProfileScreenTest.kt
+++ b/app/src/androidTest/java/com/android/periodpals/ui/profile/ProfileScreenTest.kt
@@ -1,31 +1,48 @@
 package com.android.periodpals.ui.profile
 
 import androidx.compose.ui.test.assertIsDisplayed
+import androidx.compose.ui.test.assertIsNotDisplayed
 import androidx.compose.ui.test.assertTextEquals
 import androidx.compose.ui.test.junit4.createComposeRule
 import androidx.compose.ui.test.onNodeWithTag
-import androidx.navigation.compose.rememberNavController
+import com.android.periodpals.ui.alert.AlertScreen
 import com.android.periodpals.ui.navigation.NavigationActions
+import com.android.periodpals.ui.navigation.Route
+import org.junit.Before
 import org.junit.Rule
 import org.junit.Test
+import org.mockito.Mockito.mock
+import org.mockito.Mockito.`when`
 
 class ProfileScreenTest {
 
+  private lateinit var navigationActions: NavigationActions
   @get:Rule val composeTestRule = createComposeRule()
+
+  @Before
+  fun setUp() {
+    navigationActions = mock(NavigationActions::class.java)
+
+    `when`(navigationActions.currentRoute()).thenReturn(Route.ALERT)
+
+    composeTestRule.setContent { AlertScreen(navigationActions) }
+  }
 
   @Test
   fun displayAllComponents() {
-    composeTestRule.setContent { ProfileScreen(NavigationActions(rememberNavController())) }
+    composeTestRule.setContent { ProfileScreen(navigationActions) }
     composeTestRule.onNodeWithTag("profileAvatar").assertIsDisplayed()
     composeTestRule.onNodeWithTag("profileName").assertIsDisplayed()
     composeTestRule.onNodeWithTag("Description").assertIsDisplayed()
-    composeTestRule.onNodeWithTag("reviewOne").assertIsDisplayed()
-    composeTestRule.onNodeWithTag("reviewTwo").assertIsDisplayed()
+    composeTestRule.onNodeWithTag("noReviewsCardText").assertIsDisplayed()
+    composeTestRule.onNodeWithTag("bottomNavigationMenu").assertIsDisplayed()
+    composeTestRule.onNodeWithTag("topBar").assertIsDisplayed()
+    composeTestRule.onNodeWithTag("goBackButton").assertIsNotDisplayed()
   }
 
   @Test
   fun profileScreen_hasCorrectContent() {
-    composeTestRule.setContent { ProfileScreen(NavigationActions(rememberNavController())) }
+    composeTestRule.setContent { ProfileScreen(navigationActions) }
     composeTestRule.onNodeWithTag("profileName").assertTextEquals("Name")
     composeTestRule.onNodeWithTag("Description").assertTextEquals("Description")
   }
--- a/app/src/androidTest/java/com/android/periodpals/ui/profile/ProfileScreenTest.kt
+++ b/app/src/androidTest/java/com/android/periodpals/ui/profile/ProfileScreenTest.kt
@@ -5,7 +5,6 @@ import androidx.compose.ui.test.assertIsNotDisplayed
 import androidx.compose.ui.test.assertTextEquals
 import androidx.compose.ui.test.junit4.createComposeRule
 import androidx.compose.ui.test.onNodeWithTag
-import com.android.periodpals.ui.alert.AlertScreen
 import com.android.periodpals.ui.navigation.NavigationActions
 import com.android.periodpals.ui.navigation.Route
 import org.junit.Before
@@ -23,14 +22,13 @@ class ProfileScreenTest {
   fun setUp() {
     navigationActions = mock(NavigationActions::class.java)
 
-    `when`(navigationActions.currentRoute()).thenReturn(Route.ALERT)
+    `when`(navigationActions.currentRoute()).thenReturn(Route.PROFILE)
 
-    composeTestRule.setContent { AlertScreen(navigationActions) }
+    composeTestRule.setContent { ProfileScreen(navigationActions) }
   }
 
   @Test
   fun displayAllComponents() {
-    composeTestRule.setContent { ProfileScreen(navigationActions) }
     composeTestRule.onNodeWithTag("profileAvatar").assertIsDisplayed()
     composeTestRule.onNodeWithTag("profileName").assertIsDisplayed()
     composeTestRule.onNodeWithTag("Description").assertIsDisplayed()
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/android/periodpals/ui/profile/ProfileScreen.kt
```diff
@@ -51,6 +55,9 @@ fun ProfileScreen(navigationActions: NavigationActions) {
         Uri.parse("android.resource://com.android.periodpals/${R.drawable.generic_avatar}"))
   }
 
+  // Number of interactions placeholder
+  val numberInteractions = 0
```
charliemangano, line 59: nitpick: would rename this to `interactions` for brevity

app/src/main/java/com/android/periodpals/ui/profile/ProfileScreen.kt
```diff
@@ -83,7 +92,12 @@ fun ProfileScreen(navigationActions: NavigationActions) {
           )
 
           ProfileName() // Display the user's profile name.
-          ProfileDetails() // Display additional details like description and reviews.
+
+          if (numberInteractions > 0) {
+            ProfileDetails("Number of interactions: $numberInteractions")
+          } else {
+            ProfileDetails("New user")
```
charliemangano, line 99: nitpick: I think the strings here could be extracted to constant values but there is already a cleanup task that will take care of it if you don't

charliemangano (approved):
## Summary
Good tests, nice job! We'll need to add test to check that the profile correctly fetches user profile info from the VM when the VM is done.
## Important
#### Code Quality
Consistent and very readable!
#### Functionality
I think the tests are correct and complete.
## Testing
Tests seem correct and complete. They all pass.
## Documentation
Self-documenting code (great!)
## Suggestions
None
## Steps before PR approved
LGTM, see comments for nitpicks

app/src/main/java/com/android/periodpals/ui/profile/ProfileScreen.kt
lazarinibruno: By this do you mean adding the [strings as resources](https://developer.android.com/guide/topics/resources/string-resource) in the app? Because I think it is a very good idea!

app/src/main/java/com/android/periodpals/ui/profile/ProfileScreen.kt
```diff
@@ -100,23 +114,43 @@ private fun ProfileName() {
...
       modifier = Modifier.fillMaxWidth(),
       verticalArrangement = Arrangement.spacedBy(8.dp), // Space items by 8dp vertically.
   ) {
+    // Placeholder for the user's description.
+    val description = ""
+
     // Box for the description.
     Text(
         text = "Description",
         fontSize = 20.sp,
```
lazarinibruno, line 128: Strictly speaking, there should be no hard-coded values for styling in the app. Everything should derive from the theme and style we choose. Doing so allows us to have better consistency.

app/src/main/java/com/android/periodpals/ui/profile/ProfileScreen.kt
```diff
@@ -100,23 +114,43 @@ private fun ProfileName() {
...
         modifier = Modifier.padding(vertical = 8.dp).testTag("Description"),
     )
-    ProfileInfoBox(text = "", minHeight = 100.dp, Modifier)
-    Text(text = "New user / Number of interactions", fontSize = 16.sp, color = Color(101, 116, 193))
+    ProfileInfoBox(text = description, minHeight = 100.dp, Modifier)
+    Text(text = text, fontSize = 16.sp, color = Color(101, 116, 193))
```
lazarinibruno, line 132: Same situation here: the styling should come from the theme and we should avoid hard-coding style.

app/src/androidTest/java/com/android/periodpals/ui/profile/ProfileScreenTest.kt
```diff
@@ -4,28 +4,50 @@ import androidx.compose.ui.test.assertIsDisplayed
...
+  @Test
+  fun editButtonNavigatesToEditProfileScreen() {
+    composeTestRule.onNodeWithTag("editButton").performClick()
+    verify(navigationActions).navigateTo(Screen.EDIT_PROFILE)
   }
 
   @Test
-  fun profileScreen_hasCorrectContent() {
-    composeTestRule.setContent { ProfileScreen(NavigationActions(rememberNavController())) }
+  fun profileScreenHasCorrectContent() {
```
lazarinibruno, line 50: What is the difference between this test and the test `displayAllComponents()`?

app/src/main/java/com/android/periodpals/ui/profile/ProfileScreen.kt
agonzalez-r: Yes, however matching the style to theme is another task from what I understood from yesterday stand-up meeting. The PO was very clear about separating the tasks... Because it needs to be done on all of the screens

app/src/main/java/com/android/periodpals/ui/profile/ProfileScreen.kt
```diff
@@ -83,7 +92,12 @@ fun ProfileScreen(navigationActions: NavigationActions) {
           )
 
           ProfileName() // Display the user's profile name.
```
lazarinibruno, line 94: Nitpick: I don't see the need to have two separate functions for the name of the profile and the details of the profile (the `ProfileName` composable only has a `Text` composable). Why not merge them into a single composable called, for instance, `ProfileContent` or something similar?

lazarinibruno (approved):
## Summary

Thanks for your contribution :) Overall, I really like the code. It is clear, concise and well-commented.

## Improvements

- As I said in the comments, there should be no hard-coded style in the code. The theme and style should completely derive from the `Theme.kt` and `Type.kt`. I know that we don't have a finished theme and type yet, so I understand why you hard-coded them. I see two options:
  
  - Leave it as-is, and then create a task to remove all of the hard-coded values (or maybe it is already part of #84?)
    
  - Remove the styling (even if it looks terrible), and wait until we define the Theme and Type.
    
- Also, as I commented, I think that the `ProfileName` composable is not necessary. It only contains one `Text` composable. Why not merge the `ProfileName` and `ProfileDetails` composables into a single one?
  

## Steps before merge

The improvements I mentioned are not critical, so I will accept the merge. However, I think it would be great if they were addressed. In specific, the theming. We should discuss with the group how to handle the fact that the theme is not finished yet.

app/src/main/java/com/android/periodpals/ui/profile/ProfileScreen.kt
charliemangano: Either that or just `const val`s at the top of the file. It's true that xml resources would be more adaptable later on though

charliemangano:
> Leave it as-is, and then create a task to remove all of the hard-coded values (or maybe it is already part of [Clean up code from M1 #84](https://github.com/PeriodPals/periodpals/issues/84)?)

This issue will get rid of most hardcoded values. However it will not do anything in relation with the theme, as this is a big and separate task.

app/src/main/java/com/android/periodpals/ui/profile/ProfileScreen.kt
lazarinibruno: Okay, then that is how we are managing it, cool :) I must have missed that info.

agonzalez-r:
Thank you for the review @lazarinibruno 
>   * Remove the styling (even if it looks terrible), and wait until we define the Theme and Type.

For the theme integration, as I already replied in the comments, it's a task to be done on all screens, therefore a separate task and not for this sprint. 

> * Also, as I commented, I think that the `ProfileName` composable is not necessary. It only contains one `Text` composable. Why not merge the `ProfileName` and `ProfileDetails` composables into a single one?

This is a fair point, I kept the compostables structure of the previous code. Although the composable is only a `Text`, I do think it makes sense to separate it from the `Profile Details` for modularity and readability. Actually, what I think would be best is creating another composable for the reviews (for now only the `Card`). That way every "section" of the screen would be correctly separated.

app/src/main/java/com/android/periodpals/ui/profile/ProfileScreen.kt
lazarinibruno: Okay, I will create a draft for an issue "Implement xml resources" in the _For later_ column of the Scrum board. If we find time, I think it would be elegant to implement it.

lazarinibruno:
> > * Also, as I commented, I think that the `ProfileName` composable is not necessary. It only contains one `Text` composable. Why not merge the `ProfileName` and `ProfileDetails` composables into a single one?
> 
> This is a fair point, I kept the compostables structure of the previous code. Although the composable is only a `Text`, I do think it makes sense to separate it from the `Profile Details` for modularity and readability. Actually, what I think would be best is creating another composable for the reviews (for now only the `Card`). That way every "section" of the screen would be correctly separated.

I hadn't think about it in that way. Then lets leave it as is and when we have the review system in place, we can do a composable for the reviews.

Issue #78, opened 2024-10-28
Permission request for location should popup after signin


charliemangano:
# Permission request for location should popup after SignIn
The permission request pop-up for the user's location displays when opening the app for the first time.

**To Reproduce**
Install (or delete cache if already installed) and launch the app.

**Expected behavior**
The pop-up should display only after the user has logged in/signed up succesfully

**Screenshots**
![image](https://github.com/user-attachments/assets/107d1699-2110-43d4-9e16-e8f478e113e3)

**Additional context**
We need this fixed in order to do the end-to-end tests (#48)

lazarinibruno:
The development page for android states: 

_When a feature in your app needs location access, wait until the user interacts with the feature before making the permission request. _

Therefore, we are indeed asking for user permission in the wrong moment. I think we should ask for permission when the user clicks on the **map, alert list or alert**.

charliemangano:
Duplicate of #95. Leaving both issues open as there are useful information on both.

Issue #95, opened 2024-10-30
Fix geolocation


lazarinibruno:
Currently, the location logic is inside the `MainActivity`. Every time the app launches, it asks the user for permission to use their location. This is undesirable for two reasons:

- Location access should be asked for when the user interacts with a part of the app that needs location access. For instance, when the user clicks on the map.
- When we try to launch the E2E test, it fails because the app asks for permission and the system dialog pops, but the test has no way of accepting the location access request.

Therefore, the logic for the handling of the location should be placed in a class that is in charge of it.

PR #110, opened 2024-10-31
Fix/geolocation/integrate location service

lazarinibruno:
# Integrate Location Service and Remove old code

## Description
This PR integrates the newly written location service.  By doing so, the logic for the location is contained within the `LocationServiceImpl` class. Moreover, the app requests the access to the location at the right time (namely, when clicking on the Map). It closes issues #78 and #95. 

## Changes
- Removes the location access request logic from the main (since it is now handled by `LocationServiceImpl`).
- Instantiates a `locationService` in the main activity, which is then passed as an argument to `PeriodPalsApp` and consecutively to `MapScreen`.
- Adapts the `MapScreen` to the new implementation. 
- Adapted the `MapScreenTest`
 
#### Added
None.

#### Modified
- `MainActivity.kt`
- `Map.kt`
- `MapScreenTest.kt`

#### Removed
None.

## Dependencies Added
None.

## Testing
- I adapted the `MapScreenTest` to work with the new location service.
- I did two tests from the test that checks whether the map is shown when permission is granted:
    - One for when approximate location is granted
    - One for when precise location is granted 

## Screenshots
image

Location access request in the right screen (੭*ˊᵕˋ)੭


agonzalez-r (changes requested):
## Summary
Thank you for this implementation! It contributes a lot on the Map feature of the app. 
Sadly I can't manually check if it works at expected (my emulator is only connected to wifi when I'm on campus for some reason so I can't login). But all the tests pass. 
## Important (which includes)
#### Code Quality
- I left a comment about the parameter `locationPermissionGranted` for `MapViewContainer()`, I think it should be replaced directly by `locationGrantedType` so you can directly use it 
#### Functionality
- ***Question***: where is the distinction between the precise and approximate location granted type treated? Since it's the same code for both types, I figure that it is automatically treated different by the location listener, but I'm not sure
## Testing
The tests for this screen are not complete: 
- Mainly, there is no tests to assert that components like `BottomNavigationMenu` and `TopAppBar` are correctly displayed (which also is not consistent with the rest of screen tests)
- For all tests (which difference is only the type of permission granted) you assert the same thing, only that map view exists. I think each tests should verify something more specific for each type (if possible). At least, when no permission is granted, could you verify that the default location is displayed? 


## Suggestions
- Tests more precise for different typer of granted permission types? (again, not sure if this is possible)
## Steps before PR approved
- Add tests for the correct display of `BottomNavigationMenu` and `TopAppBar` in `MapScreenTest.kt`
- Modify `MapViewContainer()` parameters for better code quality

lazarinibruno:
> * For all tests (which difference is only the type of permission granted) you assert the same thing, only that map view exists. I think each tests should verify something more specific for each type (if possible). At least, when no permission is granted, could you verify that the default location is displayed?

Well, as @taghizadlaura said in some of the meetings, there is no direct way of testing the content of the map. This is why we only assert that the map is displayed given the different permissions. Moreover, this PR focuses on the integration of the location service, not in the map itself.

app/src/main/java/com/android/periodpals/ui/map/Map.kt
```diff
@@ -33,13 +36,23 @@ private const val TAG = "MapView"
...
     navigationActions: NavigationActions
 ) {
   val context = LocalContext.current
   val fusedLocationClient = remember { LocationServices.getFusedLocationProviderClient(context) }
   val mapView = remember { MapView(context) }
 
+  locationService.requestUserPermissionForLocation()
+  val locationGrantedType = locationService.locationGrantedType.collectAsState().value
+
+  val locationPermissionGranted =
```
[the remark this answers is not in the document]
lazarinibruno: I know. I did that as a temporary fix to integrate the location service. In S4, I will finish the location service to handle what to do given the different types of location access granted #130.

agonzalez-r (approved):
Okay, we'll get to the changes later.

app/src/main/java/com/android/periodpals/ui/map/Map.kt
```diff
@@ -31,15 +34,21 @@ private val DEFAULT_LOCATION = GeoPoint(46.5191, 6.5668)
 private const val TAG = "MapView"
 
 @Composable
-fun MapScreen(
-    modifier: Modifier = Modifier,
-    locationPermissionGranted: Boolean,
-    navigationActions: NavigationActions
-) {
+fun MapScreen(locationService: LocationServiceImpl, navigationActions: NavigationActions) {
```
taghizadlaura, line 37: Problem with the declaration in the MainActivity

app/src/main/java/com/android/periodpals/MainActivity.kt
```diff
@@ -143,7 +94,7 @@ fun PeriodPalsApp(
     // Map
     navigation(startDestination = Screen.MAP, route = Route.MAP) {
       composable(Screen.MAP) {
-        MapScreen(Modifier.fillMaxSize(), locationPermissionGranted, navigationActions)
+        MapScreen(Modifier.fillMaxSize(), locationService, navigationActions)
```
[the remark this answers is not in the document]
lazarinibruno: Should be fixed now.

app/src/androidTest/java/com/android/periodpals/ui/map/MapScreenTest.kt
```diff
@@ -2,41 +2,131 @@ package com.android.periodpals.ui.map
...
+        mapOf(
+            Manifest.permission.ACCESS_FINE_LOCATION to true,
+            Manifest.permission.ACCESS_COARSE_LOCATION to true)
+
+    permissionCallbackCaptor.value.onActivityResult(permissions)
+
+    composeTestRule.setContent {
+      PeriodPalsAppTheme {
+        MapScreen(
+            modifier = Modifier,
```
taghizadlaura, line 102: Same problem as the previous one in the MainActivity

app/src/androidTest/java/com/android/periodpals/ui/map/MapScreenTest.kt
```diff
@@ -2,41 +2,131 @@ package com.android.periodpals.ui.map
...
+            Manifest.permission.ACCESS_FINE_LOCATION to false,
+            Manifest.permission.ACCESS_COARSE_LOCATION to false)
+
+    permissionCallbackCaptor.value.onActivityResult(permissions)
+
     composeTestRule.setContent {
       PeriodPalsAppTheme {
         MapScreen(
-            locationPermissionGranted = false,
+            modifier = Modifier,
```
taghizadlaura, line 123: same here

app/src/androidTest/java/com/android/periodpals/ui/map/MapScreenTest.kt
```diff
@@ -2,41 +2,131 @@ package com.android.periodpals.ui.map
...
+            Manifest.permission.ACCESS_COARSE_LOCATION to true)
+
+    // Simulate the permission grant
+    permissionCallbackCaptor.value.onActivityResult(permissions)
+
     composeTestRule.setContent {
       PeriodPalsAppTheme {
         MapScreen(
-            locationPermissionGranted = true,
+            modifier = Modifier,
```
taghizadlaura, line 80: same here

taghizadlaura (changes requested):

taghizadlaura (approved):
LGTM now, but there is a bug on the approximate location. We decided to fix it in a different task : #130

PR #138, opened 2024-11-04
Feat/alert/model: Create Alert data class and model interface

agonzalez-r:
# Create Alert data class and model interface

## Description
This PR introduces the bases for the model and view model for Alerts. 
I followed [this example](https://github.com/hieuwu/product-sample-supabase-kt/tree/main/app/src/main/java/com/example/manageproducts/data/repository) (which is a full example provided by [supabase official documentation](https://supabase.com/docs/guides/getting-started/tutorials/with-kotlin)) of model and view model using supabase as the data base, it was extremely useful. 

The data class `Alert.kt` declares all the variables needed to create an alert object. Note that the data class for objects stored in Supabase needs to be serializable. This is necessary to enable the conversion of Kotlin objects to and from formats like JSON, which is typically used for data storage and retrieval in Supabase.

Then, the interface `AlertModel.kt` declares all necessary functions to treat alerts. 

Both of these files will probably have minor modifications in the future. The purpose of this PR is mainly to check that the idea of implementation is correct, before proceeding with the implementation of the model and view model, and fix any conflicts if needed. 

#### Added
- `periodicals/model/alert/Alert.kt`: data class
- `periodicals/model/alert/AlertModel.kt`: model interface 

## Testing
No testing possible at the moment.


app/src/main/java/com/android/periodpals/model/alert/AlertModel.kt
```diff
@@ -0,0 +1,65 @@
...
+
+  /**
+   * Adds a new alert.
+   *
+   * @param alert The alert to be added.
+   * @param onSuccess Callback function to be called on successful addition.
+   * @param onFailure Callback function to be called on failure, with the exception as a parameter.
+   * @return Boolean indicating the success of the operation.
+   */
+  suspend fun addAlert(alert: Alert, onSuccess: () -> Unit, onFailure: (Exception) -> Unit): Boolean
```
coaguila, line 14: Since we are using callbacks we need to avoid that our functions return anything other than Unit, instead you should call `onSuccess`. If you want to return a boolean make `onSuccess` a `(Boolean) -> Unit`.

coaguila, line 14:
```suggestion
  suspend fun addAlert(alert: Alert, onSuccess: (Boolean) -> Unit, onFailure: (Exception) -> Unit)
```

app/src/main/java/com/android/periodpals/model/alert/AlertModel.kt
```diff
@@ -0,0 +1,65 @@
...
+   * @param onSuccess Callback function to be called on successful retrieval, with the alert as a
+   *   parameter.
+   * @param onFailure Callback function to be called on failure, with the exception as a parameter.
+   * @return The retrieved alert, or null if not found.
+   */
+  suspend fun getAlert(
+      idAlert: String,
+      onSuccess: (Alert) -> Unit,
+      onFailure: (Exception) -> Unit
+  ): Alert?
```
coaguila, line 29: Like before no need for a return type if there is a callback. Rest is correct, great job.

coaguila, line 29:
```suggestion
  /**
   * Retrieves an alert by its ID.
   *
   * @param idAlert The ID of the alert to be retrieved.
   * @param onSuccess Callback function to be called on successful retrieval, with the alert as a
   *   parameter.
   * @param onFailure Callback function to be called on failure, with the exception as a parameter.
   */
  suspend fun getAlert(
      idAlert: String,
      onSuccess: (Alert?) -> Unit,
      onFailure: (Exception) -> Unit
  )
```

app/src/main/java/com/android/periodpals/model/alert/AlertModel.kt
```diff
@@ -0,0 +1,65 @@
...
+   *
+   * @param onSuccess Callback function to be called on successful retrieval, with the list of
+   *   alerts as a parameter.
+   * @param onFailure Callback function to be called on failure, with the exception as a parameter.
+   * @return The list of retrieved alerts, or null if none found.
+   */
+  suspend fun getAlerts(
+      onSuccess: (List<Alert>) -> Unit,
+      onFailure: (Exception) -> Unit
+  ): List<Alert>?
```
coaguila, line 42:
```suggestion
  suspend fun getAlerts(
      onSuccess: (List<Alert>) -> Unit,
      onFailure: (Exception) -> Unit
  )
```

coaguila (changes requested):
There's a couple of problems with how callbacks are implemented but for the most part it is correct. Please correct this before merging! Thanks for the hard work! ( •̀ ω •́ )y

app/src/main/java/com/android/periodpals/model/alert/AlertModel.kt
agonzalez-r: Oh, I see. Thank you for the clarification I will c change that ASAP

app/src/main/java/com/android/periodpals/model/alert/AlertModel.kt
agonzalez-r: did you mean `onSuccess: (Alert) -> Unit` ?

agonzalez-r: do you think it's best to leave the Alert parameter nullable here `onSuccess: (Alert?) -> Unit` (on line 26)?

coaguila: ah wait i missunderstood, yes i meant ```onSuccess: (Alert) -> Unit```

coaguila (approved):
Lgtm! Keep up the great work!

app/src/main/java/com/android/periodpals/model/alert/Alert.kt
```diff
@@ -0,0 +1,36 @@
...
+enum class Product {
+  TAMPON,
+  PAD
+}
+
+enum class Urgency {
+  URGENT,
+  NORMAL,
+  NOT_URGENT,
+}
```
charliemangano, line 30: Good idea to have them in an `enum` class. However, I don't think these are the correct labels. Looking at the Figma, it seems that it should be high, medium and low. 

I also suggest inverting the order since `enum` classes essentially map a name to an `int`, starting from zero. What you have here would map `URGENT` to 0, `NORMAL` to 1 and `NOT_URGENT` to 2, but it would be more intuitive to have it the other way around.

```suggestion
enum class Urgency {
  LOW,
  MEDIUM,
  HIGH,
}
```

app/src/main/java/com/android/periodpals/model/alert/Alert.kt
```diff
@@ -0,0 +1,36 @@
...
+
+enum class Status {
```
charliemangano, line 32: We should discuss this with the others and modify the Figma and the architecture diagram but I think we should add a "Canceled" category. No need to implement it yet though.

app/src/main/java/com/android/periodpals/model/alert/AlertModel.kt
```diff
@@ -0,0 +1,55 @@
...
+  suspend fun getAlert(idAlert: String, onSuccess: (Alert) -> Unit, onFailure: (Exception) -> Unit)
+
+  /**
+   * Retrieves all alerts in data base.
+   *
+   * @param onSuccess Callback function to be called on successful retrieval, with the list of
+   *   alerts as a parameter.
+   * @param onFailure Callback function to be called on failure, with the exception as a parameter.
+   */
+  suspend fun getAlerts(onSuccess: (List<Alert>) -> Unit, onFailure: (Exception) -> Unit)
```
charliemangano, line 32: I suggest renaming this to something like `getAllAlerts` to avoid confusing it with `getAlert`
```suggestion
  suspend fun getAllAlerts(onSuccess: (List<Alert>) -> Unit, onFailure: (Exception) -> Unit)
```

app/src/main/java/com/android/periodpals/model/alert/Alert.kt
agonzalez-r: Yes we should discuss that tomorrow because originally we had a canceled status and other members of the team told me to remove it.

charliemangano (approved):
Perfect, thanks for the changes! LGTM

PR #143, opened 2024-11-07
Feat/alert/model impl: Create AlertModelSupabase

agonzalez-r:
# Feat/alert/model-impl: Implementation of the Alert model with supabase

## Description
This PR introduces the class implementation (`AlertModelSupabase`) of the previously done model interface (`AlertModel`), using supabase as the database. 

I have some doubts about this implementation and the model in general, some of them are: 
- do we really need to declare the parameters `id` and `uid` in the data class `Alert`? or should these values be managed directly by supabase ? 
- should we add a function `getMyAlerts(uid, ...)`, giving the `uid` as parameter and selecting all alerts linked to that uid 

#### Added
- `main/.../model/alert/AlertModelSupabase`: implementation of all the functions defined in the interface 
#### Modified
- `main/.../model/alert/AlertModelSupabase`: had to add a parameter `idAlert` in the `updateAlert()` function to be able to identify which alert to modify 

## Testing
Not possible at the moment (View Model will be tested)

### DIFF
--- a/app/src/main/java/com/android/periodpals/model/alert/Alert.kt
+++ b/app/src/main/java/com/android/periodpals/model/alert/Alert.kt
@@ -5,12 +5,12 @@ import kotlinx.serialization.Contextual
 /**
  * Data class representing an alert.
  *
- * @property id The unique identifier of the alert.
- * @property uid The user ID associated with the alert.
- * @property name The name of the alert.
+ * @property id The unique identifier of the alert, given when created in Supabase.
+ * @property uid The user ID associated with the alert
+ * @property name The name of the user who created the Alert.
  * @property product The product associated with the alert.
  * @property urgency The urgency level of the alert.
- * @property createdAt The date and time when the alert was created.
+ * @property createdAt The date and time when the alert was created, normally initialized in Supabase, otherwise declare it as : LocalDateTime(2022, 1, 1, 0, 0).toString()
  * @property location The location of the alert.
  * @property message The message associated with the alert.
  * @property status The current status of the alert.
--- a/app/src/main/java/com/android/periodpals/model/alert/Alert.kt
+++ b/app/src/main/java/com/android/periodpals/model/alert/Alert.kt
@@ -2,8 +2,6 @@ package com.android.periodpals.model.alert
 
 import java.time.LocalDateTime
 import kotlinx.serialization.Contextual
-import kotlinx.serialization.SerialName
-import kotlinx.serialization.Serializable
 
 /**
  * Data class representing an alert.
@@ -18,17 +16,16 @@ import kotlinx.serialization.Serializable
  * @property message The message associated with the alert.
  * @property status The current status of the alert.
  */
-@Serializable
 data class Alert(
-    @SerialName("id") val id: String,
-    @SerialName("uid") val uid: String,
-    @SerialName("name") val name: String,
-    @SerialName("product") val product: Product,
-    @SerialName("urgency") val urgency: Urgency,
-    @SerialName("createdAt") @Contextual val createdAt: LocalDateTime,
-    @SerialName("location") val location: String, // TODO: Create data class Location
-    @SerialName("message") val message: String,
-    @SerialName("status") val status: Status
+    val id: String? = " ", // given when created in supabase
+    val uid: String,
+    val name: String,
+    val product: Product,
+    val urgency: Urgency,
+    @Contextual val createdAt: LocalDateTime,
+    val location: String, // TODO: Create data class Location
+    val message: String,
+    val status: Status
 )
 
 /** Enum class representing the product requested with the alert. */
--- a/app/src/main/java/com/android/periodpals/model/alert/Alert.kt
+++ b/app/src/main/java/com/android/periodpals/model/alert/Alert.kt
@@ -1,7 +1,5 @@
 package com.android.periodpals.model.alert
 
-import kotlinx.serialization.Contextual
-
 /**
  * Data class representing an alert.
  *
@@ -10,7 +8,8 @@ import kotlinx.serialization.Contextual
  * @property name The name of the user who created the Alert.
  * @property product The product associated with the alert.
  * @property urgency The urgency level of the alert.
- * @property createdAt The date and time when the alert was created, normally initialized in Supabase, otherwise declare it as : LocalDateTime(2022, 1, 1, 0, 0).toString()
+ * @property createdAt The date and time when the alert was created, normally initialized in
+ *   Supabase, otherwise declare it as : LocalDateTime(2022, 1, 1, 0, 0).toString()
  * @property location The location of the alert.
  * @property message The message associated with the alert.
  * @property status The current status of the alert.
--- a/app/src/main/java/com/android/periodpals/model/alert/Alert.kt
+++ b/app/src/main/java/com/android/periodpals/model/alert/Alert.kt
@@ -21,7 +21,7 @@ data class Alert(
     val name: String,
     val product: Product,
     val urgency: Urgency,
-    @Contextual val createdAt: String,
+    val createdAt: String?,
     val location: String, // TODO: Create data class Location
     val message: String,
     val status: Status
--- a/app/src/main/java/com/android/periodpals/model/alert/Alert.kt
+++ b/app/src/main/java/com/android/periodpals/model/alert/Alert.kt
@@ -17,8 +17,8 @@ import kotlinx.serialization.Contextual
  * @property status The current status of the alert.
  */
 data class Alert(
-    val id: String? = " ", // given when created in supabase
-    val uid: String,
+    val id: String?, // given when created in supabase
+    val uid: String?,
     val name: String,
     val product: Product,
     val urgency: Urgency,
--- a/app/src/main/java/com/android/periodpals/model/alert/Alert.kt
+++ b/app/src/main/java/com/android/periodpals/model/alert/Alert.kt
@@ -1,6 +1,5 @@
 package com.android.periodpals.model.alert
 
-import java.time.LocalDateTime
 import kotlinx.serialization.Contextual
 
 /**
@@ -22,7 +21,7 @@ data class Alert(
     val name: String,
     val product: Product,
     val urgency: Urgency,
-    @Contextual val createdAt: LocalDateTime,
+    @Contextual val createdAt: String,
     val location: String, // TODO: Create data class Location
     val message: String,
     val status: Status
--- a/app/src/main/java/com/android/periodpals/model/alert/AlertModel.kt
+++ b/app/src/main/java/com/android/periodpals/model/alert/AlertModel.kt
@@ -34,17 +34,11 @@ interface AlertModel {
   /**
    * Updates an existing alert (edited).
    *
-   * @param idAlert The ID of the alert to be updated.
-   * @param alert Updated parameters for the Alert.
+   * @param alert The alert to be updated.
    * @param onSuccess Callback function to be called on successful update.
    * @param onFailure Callback function to be called on failure, with the exception as a parameter.
    */
-  suspend fun updateAlert(
-      alert: Alert,
-      idAlert: String,
-      onSuccess: () -> Unit,
-      onFailure: (Exception) -> Unit
-  )
+  suspend fun updateAlert(alert: Alert, onSuccess: () -> Unit, onFailure: (Exception) -> Unit)
 
   /**
    * Deletes an alert by its ID.
--- a/app/src/main/java/com/android/periodpals/model/alert/AlertModel.kt
+++ b/app/src/main/java/com/android/periodpals/model/alert/AlertModel.kt
@@ -7,10 +7,11 @@ interface AlertModel {
    * Adds a new alert.
    *
    * @param alert The alert to be added.
-   * @param onSuccess Callback function to be called on successful addition.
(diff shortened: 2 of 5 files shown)


app/src/main/java/com/android/periodpals/model/alert/AlertModelSupabase.kt
```diff
@@ -0,0 +1,132 @@
...
+  /**
+   * Adds a new alert to the database.
+   *
+   * @param alert The alert to be added.
+   * @param onSuccess Callback function to be called on successful addition.
+   * @param onFailure Callback function to be called on failure, with the exception as a parameter.
+   */
+  override suspend fun addAlert(
+      alert: Alert,
+      onSuccess: () -> Unit,
```
coaguila, line 30: you could return the alertId here

app/src/main/java/com/android/periodpals/model/alert/AlertModelSupabase.kt
```diff
@@ -0,0 +1,132 @@
...
+   */
+  override suspend fun getAlert(
+      idAlert: String,
+      onSuccess: (Alert) -> Unit,
+      onFailure: (Exception) -> Unit
+  ) {
+    try {
+      val result =
+          supabase.postgrest[ALERTS].select { filter { eq("id", idAlert) } }.decodeSingle<Alert>()
+      onSuccess(result)
```
coaguila, line 59: Don't forget to add `Log.d` before `onSuccess` to help debugging later on.

app/src/main/java/com/android/periodpals/model/alert/AlertModelSupabase.kt
agonzalez-r: That makes sense in the case where the idAlert is defined once it is stored in supabase but right now it is a part of the data class so the parameter `alert` that is passed to the function should already have an idAlert, possibly generated in the view model, right? 

I think it really depends on the implementation but it is one of the questions I asked in the PR description, did you take a look at them?

agonzalez-r: or maybe we can turn the parameter nullable `val id: String? = ""` so that when the alert is created, we leave this value empty and the id is created directly in supabase, but still accessible through the data class?

agonzalez-r:
# Feat/alert/model-impl: Implementation of the Alert model with supabase
## Description
This PR closes issue #117 by implementing and testing the functions declared in the interface `AlertsModel`. It also fixed some issues with the data class declaration, after further investigation about the proper methodology.

For better understanding of why some values in the `Alert` data class are nullable, check the table in supabase and their corresponding default values (some values are filled while the instance is created in supabase). 
## Changes
## Files 
#### Added
- `model/alert/AlertDto.kt`: data transfer object which is a serializable version of `Alert.kt`. This type of object is the one used for interacting with the repository, while the normal data class is used for UI usability. 

#### Modified
- `model/alert/Alert.kt`: made non-serializable, changed the type of some values (specially `createdAt` value was converted to string because of conflicts with "serializing" `LocalDateTime`)
- `model/alert/AlertModelSupabase`: ensured proper use of the `AlertDto` data class when calling repository functions and also did proper conversion to `Alert` when the functions had values to return. 
 
## Testing
- `test[UnitTest]/.../model/alert/AlertModelSupabaseTest`: tests for both success and failure (mocked) supabase responses to all the functions of the model.

app/src/main/java/com/android/periodpals/model/alert/AlertModelSupabase.kt
```diff
@@ -0,0 +1,206 @@
...
+   */
+  override suspend fun getAlert(
+      idAlert: String,
+      onSuccess: (Alert) -> Unit,
+      onFailure: (Exception) -> Unit
+  ) {
+    try {
+      val result =
+          supabase.postgrest[ALERTS]
+              .select { filter { eq("id", idAlert) } }
```
charliemangano, line 80: What is the `eq` here? Does normal structural equality `==` not work?

app/src/main/java/com/android/periodpals/model/alert/AlertModelSupabase.kt
```diff
@@ -0,0 +1,206 @@
...
+    try {
+      val result =
+          supabase.postgrest[ALERTS].select { filter { eq("uid", uid) } }.decodeList<AlertDto>()
+      Log.d(TAG, "getMyAlerts: Success")
+      onSuccess(result.map { it.toAlert() })
+    } catch (e: Exception) {
+      Log.e(TAG, "getMyAlerts: fail to retrieve alerts: ${e.message}")
+      onFailure(e)
+    }
+  }
```
charliemangano, line 133: See `AlertModel` interface for my comment on this

app/src/main/java/com/android/periodpals/model/alert/AlertModelSupabase.kt
```diff
@@ -0,0 +1,206 @@
...
+            AlertDto(
+                id = alert.id,
+                uid = alert.uid,
+                name = alert.name,
+                product = alert.product,
+                urgency = alert.urgency,
+                createdAt = alert.createdAt,
+                location = alert.location,
+                message = alert.message,
+                status = alert.status)
```
charliemangano, line 165: Here too you can use the `AlertDto` constructor.

app/src/main/java/com/android/periodpals/model/alert/AlertModelSupabase.kt
```diff
@@ -0,0 +1,206 @@
...
+        supabase.postgrest[ALERTS].update({
+          set("name", alertDto.name)
+          set("product", alertDto.product)
+          set("urgency", alertDto.urgency)
+          set("location", alertDto.location)
+          set("message", alertDto.message)
+          set("status", alertDto.status)
+        }) {
+          filter { eq("id", alert.id) }
```
charliemangano, line 174: Here as well, why the `eq`

app/src/main/java/com/android/periodpals/model/alert/AlertModelSupabase.kt
```diff
@@ -0,0 +1,206 @@
...
+   * @param onSuccess Callback function to be called on successful deletion.
+   * @param onFailure Callback function to be called on failure, with the exception as a parameter.
+   */
+  override suspend fun deleteAlertById(
+      idAlert: String,
+      onSuccess: () -> Unit,
+      onFailure: (Exception) -> Unit
+  ) {
+    try {
+      supabase.postgrest[ALERTS].delete { filter { eq("id", idAlert) } }
```
charliemangano, line 198: Again, `eq`?

app/src/test/java/com/android/periodpals/model/alert/AlertModelSupabaseTest.kt
```diff
@@ -0,0 +1,220 @@
...
+import org.junit.Assert.assertEquals
+import org.junit.Assert.fail
+import org.junit.Before
+import org.junit.Test
+
+class AlertModelSupabaseTest {
+
+  private lateinit var alertModelSupabase: AlertModelSupabase
+
+  companion object {
```
charliemangano, line 23: Weird mix of snake case and camel case ●﹏●

app/src/test/java/com/android/periodpals/model/alert/AlertModelSupabaseTest.kt
```diff
@@ -0,0 +1,220 @@
+import com.android.periodpals.model.alert.Alert
```
charliemangano, line 1: Very complete tests, and very clean code as well, good job!

charliemangano (changes requested):
**WIP sorry je vais dodo mais j'ai peur que ça s'enregistre pas mdr**

> do we really need to declare the parameters id and uid in the data class Alert? or should these values be managed directly by supabase ?

I'm leaving this question to @coaguila, he'll know much more about this.

> should we add a function getMyAlerts(uid, ...), giving the uid as parameter and selecting all alerts linked to that uid

No, this should be done in the view mode. The model should stay as agnostic as possible of what is happening in the rest of the app. However, we can have a `getAlertsFilteredBy` function taking a `(Alert) -> Boolean` lambda as parameter. This would let the view model handle the actual logic but would make the model much more usable while still staying agnostic.

good job updating the docs
make sure to delete comments and TODOs, or to explicit them. Comments should be about the why not the how, and TODOs should be converted to issues or tasks.

app/src/main/java/com/android/periodpals/model/alert/Alert.kt
```diff
@@ -1,34 +1,29 @@
...
+    val name: String,
+    val product: Product,
+    val urgency: Urgency,
+    val createdAt: String?,
+    val location: String, // TODO: Create data class Location
+    val message: String,
+    val status: Status
 )
 
 /** Enum class representing the product requested with the alert. */
```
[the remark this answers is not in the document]
agonzalez-r: Yes that's true!

app/src/main/java/com/android/periodpals/model/alert/AlertModel.kt
```diff
@@ -31,10 +32,25 @@ interface AlertModel {
...
+      uid: String,
+      onSuccess: (List<Alert>) -> Unit,
+      onFailure: (Exception) -> Unit
+  )
+
   /**
    * Updates an existing alert (edited).
    *
-   * @param alert The alert to be updated.
+   * @param alert Updated parameters for the Alert, id must not be null to know which alert to
```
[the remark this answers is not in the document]
agonzalez-r: I wanted to specify that it is the **alert with the updated parameters** that you need to pass as argument. And since to find which alert it is we need its `id`, we need that to be non-null inside the Alert. 
Could you propose a better way of describing this?

app/src/main/java/com/android/periodpals/model/alert/AlertModelSupabase.kt
agonzalez-r: That's right! forgot to use it

app/src/main/java/com/android/periodpals/model/alert/AlertModelSupabase.kt
agonzalez-r: because we need to apply that only on that specific alert (with `id` == `alert.id`). Without the `eq` it applies that to all alerts in the table

app/src/main/java/com/android/periodpals/model/alert/AlertModelSupabase.kt
agonzalez-r: It's the only way to specify which alert.

app/src/test/java/com/android/periodpals/model/alert/AlertModelSupabaseTest.kt
agonzalez-r: Yes I followed it to be consistent with the other model unit tests :( like `AuthenticationModelSupabaseTest`, I just realised it's not the case for `UserModelSupabaseTest`

app/src/main/java/com/android/periodpals/model/alert/AlertModel.kt
```diff
@@ -31,10 +32,25 @@ interface AlertModel {
...
+   * @param uid The UID of the user whose alerts are to be retrieved.
+   * @param onSuccess Callback function to be called on successful retrieval, with the list of
+   *   alerts as a parameter.
+   * @param onFailure Callback function to be called on failure, with the exception as a parameter.
+   */
+  suspend fun getMyAlerts(
+      uid: String,
+      onSuccess: (List<Alert>) -> Unit,
+      onFailure: (Exception) -> Unit
+  )
```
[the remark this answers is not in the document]
agonzalez-r: Okay, I can change that !

app/src/main/java/com/android/periodpals/model/alert/AlertModelSupabase.kt
charliemangano: Oh yes you're right, I finally found the docs for it this morning

app/src/main/java/com/android/periodpals/model/alert/AlertModel.kt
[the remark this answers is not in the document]
charliemangano: I'd say something along the lines of:
```suggestion
   * @param alert Alert with updated parameters. `id` must not be null to know which alert to
```

coaguila (approved):
Lgtm

app/src/main/java/com/android/periodpals/model/alert/AlertModelSupabase.kt
```diff
@@ -0,0 +1,186 @@
...
+  ) {
+    try {
+      val result = supabase.postgrest[ALERTS].select { filter(cond) }.decodeList<AlertDto>()
+      Log.d(TAG, "getMyAlerts: Success")
+      onSuccess(result.map { it.toAlert() })
+    } catch (e: Exception) {
+      Log.e(TAG, "getMyAlerts: fail to retrieve alerts: ${e.message}")
+      onFailure(e)
+    }
+  }
```
charliemangano, line 124: Good generalisation of old `getMyAlerts` into `getAlertsFilteredBy`, thanks for implementing my comment suggestion!

charliemangano (approved):
LGTM! Thanks for implementing the points I had highlighted.

Issue #148, opened 2024-11-07
Change Map Services files names to GPS


taghizadlaura:
For this task we only want to create a simple GPSLocation class to make the change to MapLibre easier, and rename LocationServiceImpl to GPSServiceImpl.kt (same for LocationService.kt)

PR #149, opened 2024-11-07
Feat/geolocation/change map services names to gps

taghizadlaura:
# Refactor Location Services to Support MapLibre Integration

## Description  
This PR introduces a new `GPSLocation` class to simplify working with GPS data and make the transition to MapLibre smoother. Additionally, it refactors and renames `LocationServiceImpl` to `GPSServiceImpl.kt` for clarity, and similarly renames the interface `LocationService.kt` to `GPSService.kt`. This restructuring lays the groundwork for future updates that involve MapLibre.

It closes issue #148.

## Changes  
- Created a new `GPSLocation` class to encapsulate GPS-related data.
- Renamed `LocationServiceImpl` to `GPSServiceImpl.kt` to better reflect its role in handling GPS data.
- Renamed `LocationService.kt` to `GPSService.kt` to align with the new naming convention.
  
#### Added
- `GPSLocation.kt` – New class to represent a GPS location with fields like latitude, longitude, and accuracy.
  
#### Modified  
- `GPSServiceImpl.kt` – Refactored to implement the renamed `GPSService.kt` interface and use the new `GPSLocation` class.
- `GPSService.kt` – Interface renamed to align with the new GPS service class and related functionalities.

## Testing  
- Unit tests have been updated to reflect the new class names and structure. All tests have been run successfully to ensure the changes don't break existing functionality.

### DIFF
--- a/app/src/main/java/com/android/periodpals/model/location/GPSLocation.kt
+++ b/app/src/main/java/com/android/periodpals/model/location/GPSLocation.kt
@@ -1,0 +1,17 @@
+package com.android.periodpals.model.location
+
+import org.osmdroid.util.GeoPoint
+
+data class GPSLocation(val lat: Double, val long: Double) {
+
+    /**
+     * Transform this location into a [GeoPoint].
+     *
+     * @return The respective [GeoPoint].
+     */
+    fun toGeoPoint() = GeoPoint(lat, long)
+
+    companion object {
+        val DEFAULT_LOCATION = GPSLocation(46.5191, 6.5668)
+    }
+}
--- a/app/src/main/java/com/android/periodpals/model/location/GPSLocation.kt
+++ b/app/src/main/java/com/android/periodpals/model/location/GPSLocation.kt
@@ -4,14 +4,14 @@ import org.osmdroid.util.GeoPoint
 
 data class GPSLocation(val lat: Double, val long: Double) {
 
-    /**
-     * Transform this location into a [GeoPoint].
-     *
-     * @return The respective [GeoPoint].
-     */
-    fun toGeoPoint() = GeoPoint(lat, long)
+  /**
+   * Transform this location into a [GeoPoint].
+   *
+   * @return The respective [GeoPoint].
+   */
+  fun toGeoPoint() = GeoPoint(lat, long)
 
-    companion object {
-        val DEFAULT_LOCATION = GPSLocation(46.5191, 6.5668)
-    }
+  companion object {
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/android/periodpals/model/location/GPSLocation.kt
```diff
@@ -0,0 +1,17 @@
...
+  /**
+   * Transform this location into a [GeoPoint].
+   *
+   * @return The respective [GeoPoint].
+   */
+  fun toGeoPoint() = GeoPoint(lat, long)
+
+  companion object {
+    val DEFAULT_LOCATION = GPSLocation(46.5191, 6.5668)
+  }
```
francelu, line 16: What does companion object add in this context of a data class ? Would private val ou just val not be enough ?

francelu (approved):
## Summary
LGTM. The clean-up of the code is a beneficial step towards maintaining a clean and consistent codebase. Nice work!

## Important

#### Code Quality
- The overall readability and consistency have been significantly improved due to the code clean-up.

#### Functionality
- **Clarification needed**: In the file review, I raised a question about the use of a `companion object`. I didn't clearly understand why it was necessary in this particular context. It would be helpful if you could explain shortly.

## Testing
- It appears that all existing tests pass, and the new code modifications did not break any functionality.

## Documentation
- Nothing to suggest.

## Suggestions
- None. Reply to my question in file review.

## Steps before PR approved
- Clarify regarding the reason for the `companion object` and not just `val`.

lazarinibruno (approved):
- The new name is better!
- The Location Data Class will be very useful when finishing to implement the LocationServiceImpl. 

LGTM

app/src/main/java/com/android/periodpals/model/location/GPSLocation.kt
taghizadlaura: So, in this case, we’re using the companion object to define DEFAULT_LOCATION as a constant that belongs specifically to the GPSLocation class. By putting it in the companion object, we’re saying, "This is tied to GPSLocation," and we can access it like GPSLocation.DEFAULT_LOCATION. It also gives us the benefit of being able to group it with other class-related logic later, making it more modular and easier to maintain.
If we just used a top-level val outside the class, it wouldn’t be as clear that this value is meant to be part of GPSLocation, it’d just be a global constant. Using the companion object helps keep everything more organized and aligns with good object-oriented principles.

Issue #160, opened 2024-11-09
Integrate profile VM into Profile screen


charliemangano:
### Description:
This is a subtask from #154. The goal of this task is to have the user be able to see the data they entered while creating or editing their profile in the Profile screen by syncing the screen with Supabase.

### Done:
- user data (name, pfp, description) is successfully fetched and displayed from Supabase when arriving on the screen
- coverage > 80%

Issue #154, opened 2024-11-08
Integrate profile VM into profile-related screens


charliemangano:
### Description:
The goal of this task is to have the user be able to input their data into the CreateProfile or EditProfile screen, and for it to be sent back and stored in Supabase. That data should then be fetched when viewing the Profile.

### Done:
- **Functional Requirements:** 
  - user data is successfully pushed to the repository when clicking "Save" in the CreateProfile screen
  - user data is successfully updated in the repository when clicking "Save" in the EditProfile screen
  - user data is successfully fetched from the repository when view the Profile screen
- **Quality Standards:** 
  - clean code
  - documentation
  - user should see a popup if an error occurred at any point 
- **Testing Requirements:** 
  - integration tests to check that the values pushed, updated and fetched are correct
  - end-to-end tests to check that the values are shown and updated properly on the view
  - 80% coverage

### Preconditions:
Profile VM and M should be finished and working before starting this task.  
EditProfile screen should be finished before starting working on it.

### Steps to Completion:
- [x] #155, #159 
- [x] #160, #162 
- [x] #163, #192

### Points to Consider / Known Risks:
If profile VM and M is not finished, I can't get to work.

PR #162, opened 2024-11-09
feat/profile/integrate-vm-profile : Integrate profile VM into Profile screen

charliemangano:
# Integrate profile VM into Profile screen

## Description
This PR introduces syncing between our database and the Profile screen. Now, users can see the data they entered when singing up/editing their profile. It closes issue #160 and part of #154.

## Changes
Replaced placeholder values for profile picture, name and description in `Profile` screen with data fetched from the database using `UserViewModel.loadUser()` and `UserViewModel.user`. In case of the view model failing to retrieve the user's data, display "Error loading X" instead.

Added a `UserViewModel` parameter in the `ProfileScreen` that is passed from the `MainActivity`.

Fixed a test tag that was on the wrong component.

#### Added
None
#### Modified
- `app/src/main/java/com/android/periodpals/ui/profile/ProfileScreen.kt`: added syncing with database
- `app/src/main/java/com/android/periodpals/MainActivity.kt`: passed view model parameter in call to `ProfileScreen`
- `app/src/androidTest/java/com/android/periodpals/ui/profile/ProfileScreenTest.kt`: updated tests

#### Removed
None

## Dependencies Added
None

## Testing
Updated the existing tests to mock the user view model. They check the values displayed against those returned by the view model, and test for the view model failure case.
**Edit:** as discussed in the comments with @agonzalez-r, I could not find a way to check the behavior of the profile picture. The `GlideImage` component seems to still be in development so maybe there is actually no way to do it yet.


app/src/main/java/com/android/periodpals/ui/profile/ProfileScreen.kt
```diff
@@ -77,62 +83,61 @@ fun ProfileScreen(navigationActions: NavigationActions) {
...
     // Box for the description.
-    Text(
-        text = "Description",
-        fontSize = 20.sp,
-        modifier = Modifier.padding(vertical = 8.dp).testTag(ProfileScreen.DESCRIPTION_FIELD),
+    Text(text = "Description", fontSize = 20.sp, modifier = Modifier.padding(vertical = 8.dp))
+    ProfileInfoBox(
+        text = description,
+        minHeight = 100.dp,
+        Modifier,
```
agonzalez-r, line 138: Is this modifier empty? In that case why don't juste remove it

app/src/main/java/com/android/periodpals/ui/profile/ProfileScreen.kt
```diff
@@ -158,7 +163,7 @@ private fun ProfileDetails(text: String) {
 }
 
 @Composable
-private fun ProfileInfoBox(text: String, minHeight: Dp, modifier: Modifier) {
+private fun ProfileInfoBox(text: String, minHeight: Dp, modifier: Modifier, testTag: String) {
```
agonzalez-r, line 166: maybe we should rename this composable to `DescriptionBox`, if it's just used for that. And no need to pass the test tag as argument

charliemangano: Yeah, I've noticed as well, but @francelu is gonna be refactoring the profile screens anyway so I decided to not touch it.

app/src/main/java/com/android/periodpals/ui/profile/ProfileScreen.kt
```diff
@@ -77,62 +83,61 @@ fun ProfileScreen(navigationActions: NavigationActions) {
...
+      } else {
+        ProfileDetails("New user", userState)
+      }
+    }
+  }
 }
 
 @Composable
-private fun ProfileName() {
+private fun ProfileName(userState: State<User?>) {
```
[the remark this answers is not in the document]
agonzalez-r: but this way it's probably cleaner nvm...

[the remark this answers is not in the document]
charliemangano: I thought passing only the string might make the `ProfileName` component not recompose on change, as it is not a `State`, but that might stem from a misunderstanding of `@Composable`s on my part. I'll try and let you know.

app/src/androidTest/java/com/android/periodpals/ui/profile/ProfileScreenTest.kt
```diff
@@ -50,13 +68,33 @@ class ProfileScreenTest {
...
 
   @Test
-  fun profileScreenHasCorrectContent() {
-    composeTestRule.onNodeWithTag(ProfileScreen.NAME_FIELD).assertTextEquals("Name")
-    composeTestRule.onNodeWithTag(ProfileScreen.DESCRIPTION_FIELD).assertTextEquals("Description")
+  fun profileScreenHasCorrectContentVMSuccess() {
+    `when`(userViewModel.user).thenReturn(userState)
+
+    composeTestRule.setContent { ProfileScreen(userViewModel, navigationActions) }
+
```
agonzalez-r, line 84: Is there not a way to verify that the correct image is displayed? like comparing URLs

agonzalez-r (approved):
## Summary
Good job integrating the user view model to `ProfileScreen`! Thank you for your work
#### Code Quality
Clean code. LGTM
#### Functionality
- One minor remark about when to pass a whole `User` object to private `@Composable` functions 
## Testing
- Overall good and complete tests. All tests pass
- Missing test to assert the profile picture (`imageURL`) is correctly displayed in `ProfileScreen`
## Suggestions
- If possible, add a test to check the display of the correct profile picture

app/src/androidTest/java/com/android/periodpals/ui/profile/ProfileScreenTest.kt
charliemangano: Sadly, I could not find a way to access the `GlideImage` URL and am generally confused about how to go about testing this component. I noticed the annotation `@ExperimentalGlideComposeApi` had to be added to the file so it's still in development. I have tested the display of the image by hand, just going through the app and it seems to be working properly. We should however come back to it once we discover how to test it.

app/src/main/java/com/android/periodpals/ui/profile/ProfileScreen.kt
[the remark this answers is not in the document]
charliemangano: Update: it works just as well! I'm committing the change right now. That said the `ProfileName` component is just a text field, I believe it will get refactored when @francelu will fit this screen to the Figma

app/src/main/java/com/android/periodpals/ui/profile/ProfileScreen.kt
```diff
@@ -48,14 +47,21 @@ import com.bumptech.glide.integration.compose.ExperimentalGlideComposeApi
...
-    mutableStateOf<Uri?>(
-        Uri.parse("android.resource://com.android.periodpals/${R.drawable.generic_avatar}"))
+fun ProfileScreen(userViewModel: UserViewModel, navigationActions: NavigationActions) {
+  val context = LocalContext.current
+
+  Log.d(TAG, "Loading user data")
+  userViewModel.loadUser()
+  val userState = userViewModel.user
+  if (userState.value == null) {
+    Log.d(TAG, "User data is null")
```
lazarinibruno, line 64: **Important**: I think it would be also a good idea to notify the user with a Toast that something went wrong when loading their data.

app/src/main/java/com/android/periodpals/ui/profile/ProfileScreen.kt
```diff
@@ -48,14 +47,21 @@ import com.bumptech.glide.integration.compose.ExperimentalGlideComposeApi
...
   }
 
   // Number of interactions placeholder
```
lazarinibruno, line 67: **Note**: It would be a good idea to add the number of interactions to the model of the user, as it is a requirement for M2. Maybe we can discuss this after the next stand-up meeting.

app/src/main/java/com/android/periodpals/ui/profile/ProfileScreen.kt
```diff
@@ -77,62 +83,61 @@ fun ProfileScreen(navigationActions: NavigationActions) {
...
-          )
+  ) { padding ->
+    Column(
+        modifier = Modifier.padding(padding).padding(40.dp),
+        verticalArrangement = Arrangement.spacedBy(16.dp),
+        horizontalAlignment = Alignment.CenterHorizontally,
+    ) {
+      // Display the user's profile image.
+      GlideImage(
+          model = userState.value?.imageUrl,
```
lazarinibruno, line 94: **Question**: What happens if the `imageUrl` is null?  Is there a default image that shows?

lazarinibruno (approved):
## Summary
Thank you for the contribution 🔥🥶. I'm really happy to see that we have finally a screen of the app that successfully interacts with the backend!
 
## Important

### Code Quality
The code is readable and concise. Comments are used in a pertinent way.

### Functionality
More than anything I'm a bit worried that we are not communicating enough to the user when there is a failure fetching their data. As I commented, using Toasts would help.

Moreover, if there is a failure, I think that there should be a way to propose some steps to the user on what to do. Maybe we don't yet have the mechanisms to solve this problem (for instance, one such mechanism would be swiping down to reload the page), but we could show a [dialog](https://developer.android.com/develop/ui/compose/components/dialog) explaining that there was an error, and eventually what steps the user can take to try and solve it.

## Testing
Tests are complete:
- Verification that navigation works inside the screen ✅
- Verification that all of the components are displayed ✅
- Verification for both when the VM succeeds and fails fetching the user data ✅

## Documentation
I don't know if it is common practice to do so for composables, but maybe adding a brief documentation of what every composable does would be a good idea? Or maybe not, as the names  of the composables explain all that needs explanation? What do you think?

## Suggestions
Please read the comments I wrote.

## Steps before PR approved
- In my opinion, the most important part is properly informing the user that there was an error, and indicating them what steps they can follow to try and solve them.
- The documentation part I mentioned is not a priority for M2.
- As this changes aren't critical to the functionality of the app, I'm approving the changes. However, please do consider what I mentioned of informing the user. If you want me to take a final look before merging the PR, lmk 🤠.

app/src/main/java/com/android/periodpals/ui/profile/ProfileScreen.kt
charliemangano: Yes, I'll bring it up on Monday

app/src/main/java/com/android/periodpals/ui/profile/ProfileScreen.kt
charliemangano: I'll add a `Toast`! However, there isn't really anything that the user could do to fix this problem apart from trying again at another time, so I'll just make it display "Try again later" or something along those lines. 
Regarding your suggestion to add a dialog of some kind, I think it should be a shared component to every screen so that the UX is consistent everywhere in the app. I'll bring it up on Monday and we can see to implement it in a later sprint.

app/src/main/java/com/android/periodpals/ui/profile/ProfileScreen.kt
charliemangano: Good catch again, I'll make it be a generic avatar if the url is null

charliemangano:
> I don't know if it is common practice to do so for composables, but maybe adding a brief documentation of what every composable does would be a good idea? Or maybe not, as the names of the composables explain all that needs explanation? What do you think?

I agree! It's a bit outside of the scope of this PR but I'll do it as it's just docs.

charliemangano:
@lazarinibruno I re-requested your review as merging from main introduced quite consequent changes in the code

app/src/main/java/com/android/periodpals/ui/profile/ProfileScreen.kt
lazarinibruno: Perfect! Sounds good to me.

lazarinibruno:
> @lazarinibruno
> 
> > I don't know if it is common practice to do so for composables, but maybe adding a brief documentation of what every composable does would be a good idea? Or maybe not, as the names of the composables explain all that needs explanation? What do you think?
> 
> I agree! It's a bit outside of the scope of this PR but I'll do it as it's just docs.

True! It does not really fit into this PR's goal. Thank you anyways for taking the time and doing it.

lazarinibruno (approved):
Thanks for implementing the changes I requested. LGTM now :)

PR #173, opened 2024-11-11
Fix/style/layout responsivness/alert lists

francelu:
# Responsive layout and alert lists items

## Description
This PR introduces a responsive layout for `AlertListsScreen`. It closes issue #166, subtask from #82. 
This PR also covers the UI design of "My Alerts" and "Pals' Alerts" items individually.

## Changes
- New: UI cards for "My Alerts" and "Pals' Alerts". `AlertListsScreenTest` has been updated to cover the new code.
- Padding, shapes and typographies have been updated for `AlertListsScreen`.

#### Added
None.

#### Modified
- **`AlertListsScreenTest.kt`**: updated and new test to cover the UI and tabs functionality.
- **`C.kt`**: updated test tags.
- **`AlertLists.kt`**: new UI components for the lists' alerts, responsive layout changes.
- **`Dimens.kt`**: `iconSizeSmall` for alert cards, `cardRounded` for uniform round edges on cards.
- **`Type.kt`**: `headlineSmall` for tabs text.

#### Removed
None.

## Dependencies Added
None.

## Testing
- Updated tests for UI components and tabs functionality.
-  Verified on multiple emulators to ensure consistent and responsive design across different screen sizes.

### My Alerts Table

| Screen              | Before                              | After                               |
|---------------------|-------------------------------------|-------------------------------------|
| **Compact S**       | image | image |
| **Compact M**       | image | image |
| **Compact L**       | image | image |
| **Medium (horizontal)** | image | image |
| **Medium**          | image | image |
| **Extended**        | image | image |

### Pals Alerts Table
| Screen              | Before                              | After                               |
|---------------------|-------------------------------------|-------------------------------------|
| **Compact S**       | image |                          image |
| **Compact M**       | image | image |
| **Compact L**       | image | image |
| **Medium (horizontal)** | image | image|
| **Medium**          | image | image |
| **Extended**        | image | image |

| My Alerts | Pals Alerts |
|-----------|-------------|
| image | image |


francelu:
I tried to center the `NoAlertDialog` card. However, it didn't seem that bad to have it at the top.  
The screenshots are mostly the `NoAlertDialog` card that is displayed. Because in reality, by default there are no alerts (how I implemented it for now). At the bottom of my PR description, you'll find two screenshots where you can see the `MyAlertItem` card and the `PalAlertItem` card.

francelu:
If this PR is overwhelming and too big (which I admit), I can try to make it into two or three PRs.

app/src/main/java/com/android/periodpals/ui/alert/AlertLists.kt
```diff
@@ -88,112 +133,368 @@ fun AlertListsScreen(navigationActions: NavigationActions) {
...
+@Composable
+private fun AlertProfilePicture(alertId: String) {
+  Icon(
+      imageVector = Icons.Outlined.AccountCircle,
+      contentDescription = "Profile picture",
+      modifier =
+          Modifier.size(MaterialTheme.dimens.iconSize)
+              .wrapContentSize()
+              .testTag(AlertListsScreen.ALERT_PROFILE_PICTURE + alertId),
+  )
```
charliemangano, line 342: This should be an image (fetched from the userVM, not in the scope of this PR) and not an icon. Can you add a TODO explaining that so we don't forget

app/src/main/java/com/android/periodpals/ui/alert/AlertLists.kt
```diff
@@ -88,112 +133,368 @@ fun AlertListsScreen(navigationActions: NavigationActions) {
...
 
 /**
  * Composable function that displays an individual alert item. It includes details such as profile
- * picture, name, time, location, product type, and urgency. For now, it only displays
- * **placeholder** details.
+ * picture, time, location, name, message, product type, and urgency.
  */
 @Composable
-fun AlertItem() {
+fun PalAlertItem(alert: Alert, context: Context, isClicked: Boolean, onClick: (Boolean) -> Unit) {
```
charliemangano, line 251: The docs here are incomplete, please add the `@param`

app/src/main/java/com/android/periodpals/ui/alert/AlertLists.kt
```diff
@@ -88,112 +133,368 @@ fun AlertListsScreen(navigationActions: NavigationActions) {
...
-          modifier = Modifier.padding(horizontal = 4.dp).testTag(AlertListsScreen.ALERT_URGENCY),
-      )
     }
   }
 }
 
-/** Composable function that displays a dialog indicating that there are no alerts. */
+/** Composable function that displays the profile picture of an alert. */
+@Composable
+private fun AlertProfilePicture(alertId: String) {
```
charliemangano, line 334: The docs here are incomplete, please add the `@param`

app/src/main/java/com/android/periodpals/ui/alert/AlertLists.kt
```diff
@@ -88,112 +133,368 @@ fun AlertListsScreen(navigationActions: NavigationActions) {
...
+  )
+}
+
+/**
+ * Composable function that displays the time and location of an alert.
+ *
+ * @param alert The alert to be displayed.
+ */
+@Composable
+private fun AlertTimeAndLocation(alert: Alert, alertId: String) {
```
charliemangano, line 351: The docs here are incomplete, please add the missing `@param`

app/src/main/java/com/android/periodpals/ui/alert/AlertLists.kt
```diff
@@ -88,112 +133,368 @@ fun AlertListsScreen(navigationActions: NavigationActions) {
...
+      modifier =
+          Modifier.fillMaxWidth()
+              .wrapContentHeight()
+              .testTag(AlertListsScreen.ALERT_TIME_AND_LOCATION + alertId),
+  )
+}
+
+/** Composable function that displays the product type and urgency of an alert. */
+@Composable
+private fun AlertProductAndUrgency(alertId: String) {
```
charliemangano, line 369: The docs here are incomplete, please add the missing `@param`

app/src/main/java/com/android/periodpals/ui/alert/AlertLists.kt
```diff
@@ -88,112 +133,368 @@ fun AlertListsScreen(navigationActions: NavigationActions) {
...
+ *
+ * @param context The context to display the toast messages.
+ */
 @Composable
-fun NoAlertDialog() {
-  Column(
-      modifier = Modifier.fillMaxSize(),
-      horizontalAlignment = Alignment.CenterHorizontally,
-      verticalArrangement = Arrangement.Center,
+private fun AlertAcceptButtons(context: Context, alertId: String) {
```
charliemangano, line 402: The docs here are incomplete, please add the missing `@param`

app/src/main/java/com/android/periodpals/ui/alert/AlertLists.kt
```diff
@@ -88,112 +133,368 @@ fun AlertListsScreen(navigationActions: NavigationActions) {
...
+}
+
+/**
+ * Composable function that displays the time and location of an alert.
+ *
+ * @param alert The alert to be displayed.
+ */
+@Composable
+private fun AlertTimeAndLocation(alert: Alert, alertId: String) {
+  val formatter = DateTimeFormatter.ofPattern("HH:mm")
```
charliemangano, line 352: The date pattern could be extracted to a constant.

charliemangano (approved):
## Summary
Very good job, UI in the Alert Lists screen is now more conform to Figma and has better responsiveness. 
## Important
#### Code Quality
I really like that you refactored this file, it's really easy to read and understand now! Code is really consistent with the rest of the codebase, good job!
#### Functionality
Achieves what it needed to, nicely done!
As you mentionned, the "No alerts yet" cards are not vertically centered. This is not urgent but it might be fixed by making the alignment conditional on the emptiness of the lists.
## Testing
All tests pass and are correctly updated to account for the changes you made. I ran the app with different lists of `My Alerts` and `Pals Alerts` and the screen responds correclty.
## Documentation
Some docs are not updated, see comments. Otherwise, nice job, the rest of it is good.
## Suggestions
See comments.
## Steps before PR approved
No major problems, LGTM. Please take a look at the comments before merging.

app/src/main/java/com/android/periodpals/ui/alert/AlertLists.kt
```diff
@@ -88,112 +133,368 @@ fun AlertListsScreen(navigationActions: NavigationActions) {
...
 /**
  * Composable function that displays an individual alert item. It includes details such as profile
- * picture, name, time, location, product type, and urgency. For now, it only displays
- * **placeholder** details.
+ * picture, time, location, name, message, product type, and urgency.
  */
 @Composable
-fun AlertItem() {
+fun PalAlertItem(alert: Alert, context: Context, isClicked: Boolean, onClick: (Boolean) -> Unit) {
+  val alertId = alert.id.toString()
```
agonzalez-r, line 252: alert.id is already type String

app/src/main/java/com/android/periodpals/ui/alert/AlertLists.kt
agonzalez-r, line 334: change this parameter to `alert.uid`, to have the uid associated to the owner of the alert and possibly get data from that

app/src/androidTest/java/com/android/periodpals/ui/alert/AlertListsScreenTest.kt
```diff
@@ -56,83 +116,208 @@ class AlertListsScreenTest {
...
+    composeTestRule.onNodeWithTag(AlertListsScreen.NO_ALERTS_CARD).assertIsNotDisplayed()
+
+    MY_ALERTS_LIST.forEach { alert ->
+      val alertId: String = alert.id.toString()
+      composeTestRule
+          .onNodeWithTag(MyAlertItem.MY_ALERT + alertId)
+          .assertIsDisplayed()
+          .assertHasNoClickAction()
+      composeTestRule
+          .onNodeWithTag(AlertListsScreen.ALERT_PROFILE_PICTURE + alertId, useUnmergedTree = true)
```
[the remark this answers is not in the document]
francelu: The layout of `MyAlertItem` and `PalAlertItem` is quite complex (see below).  

So some components may overlap with others which made them not necessarily `isDisplayed()`.  
*Multiple components where taken as one big one which made the individual smaller ones untestable.* 
While with the this `useUnmergedTree = true`, I was able to test smaller components individually.  

`MyAlerItem`:
```kotlin
Card {
  Row {
    // Profile picture
    Column {
      // Time and location
      Row {
        // Product and urgency
      }
    }
    Button {
      Row {...}
    }
  }
}
``` 

And `PalAlertItem`:
```kotlin
Card {
  Coilumn {
    Row {
      // Profile picture
      Column {
        // Time and location
        // Name
        // Maybe message
      }
      Row { ... }
    }
    // Maybe
    HorizontalDivider {...}
    Row {
      Button {
        Row { ... }
      }
      Button {
        Row { ... }
      }
    }
  }
}
```

app/src/main/java/com/android/periodpals/ui/alert/AlertLists.kt
```diff
@@ -26,19 +40,36 @@ import androidx.compose.runtime.remember
...
 private const val PALS_ALERTS_TAB_TITLE = "Pals Alerts"
-private const val NO_ALERTS_DIALOG_TEXT = "No alerts here for the moment..."
+private const val NO_MY_ALERTS_DIALOG = "You haven't asked for help yet !"
+private const val NO_PAL_ALERTS_DIALOG = "No pal needs help yet !"
+private const val MY_ALERT_EDIT_TEXT = "Edit"
+private const val PAL_ALERT_ACCEPT_TEXT = "Accept"
+private const val PAL_ALERT_DECLINE_TEXT = "Decline"
+private const val TO_IMPLEMENT_EDIT_ALERT = "To implement edit alert screen"
+private const val TO_IMPLEMENT_ACCEPT_ALERT = "To implement accept alert action"
+private const val TO_IMPLEMENT_DECLINE_ALERT = "To implement decline alert action"
```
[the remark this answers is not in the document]
francelu: Applied changes !

app/src/main/java/com/android/periodpals/ui/alert/AlertLists.kt
```diff
@@ -53,29 +84,43 @@ private enum class AlertListsTab {
  * @param navigationActions The navigation actions for handling navigation events.
  */
 @Composable
-fun AlertListsScreen(navigationActions: NavigationActions) {
+fun AlertListsScreen(
+    navigationActions: NavigationActions,
+    myAlertsList: List<Alert> = emptyList(),
+    palsAlertsList: List<Alert> = emptyList(),
```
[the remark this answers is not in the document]
francelu: Yes, it was for testing purposes as I could get my mind around how to set the lists in the tests.  
Updated the documentation.

app/src/main/java/com/android/periodpals/ui/alert/AlertLists.kt
```diff
@@ -53,29 +84,43 @@ private enum class AlertListsTab {
...
+) {
+  val context = LocalContext.current
```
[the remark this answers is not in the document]
francelu: Good suggestion !  
Moved call into `MyAlertItem` and `AlertAcceptButton`.

app/src/main/java/com/android/periodpals/ui/alert/AlertLists.kt
```diff
@@ -53,29 +84,43 @@ private enum class AlertListsTab {
...
   var selectedTab by remember { mutableStateOf(SELECTED_TAB_DEFAULT) }
+  var clickedStates by remember { mutableStateOf(palsAlertsList.associate { it.id to false }) }
```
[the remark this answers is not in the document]
francelu: This is the list of states that remember for each Pals' item if it is toggled or not. I thought I needed it because else all cards would be toggled and "untoggled" at the same time, or have another trouble.  

After thoughts and recoding, I figured out an easier way to handle it, which I had at the very beginning. At that time, I had a bad understanding of what I was coding, notably with the :
```kotlin
if (isClicked && alert.status == Status.CREATED)
```

Please have a look at the new implementation.

app/src/main/java/com/android/periodpals/ui/alert/AlertLists.kt
francelu: I know...
It's done ! :))

app/src/main/java/com/android/periodpals/ui/alert/AlertLists.kt
francelu: Updated !

app/src/main/java/com/android/periodpals/ui/alert/AlertLists.kt
francelu: Updated !

app/src/main/java/com/android/periodpals/ui/alert/AlertLists.kt
francelu: Updated !

app/src/main/java/com/android/periodpals/ui/alert/AlertLists.kt
francelu: Updated !

app/src/main/java/com/android/periodpals/ui/alert/AlertLists.kt
francelu: Updated !

app/src/main/java/com/android/periodpals/ui/alert/AlertLists.kt
francelu: Extracted !

app/src/main/java/com/android/periodpals/ui/alert/AlertLists.kt
francelu: It isn't, but a `String?` which isn't necessarily a `String`.

app/src/main/java/com/android/periodpals/ui/alert/AlertLists.kt
francelu: Check the code, this `alertId` is used as an extra test tag for the `Modifier`s. If instead you use a `uid`, you might end up with two alerts from the same pal and  have troubles with the tests which will tell that "has ... children and ... siblings".  
However, I'll rename the parameter as `idTestTag` for better clarity of its purpose.

app/src/main/java/com/android/periodpals/ui/alert/AlertLists.kt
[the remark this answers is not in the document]
francelu: The idea is that in `MyAlertItem` and `PalAlertItem`, right at the beginning, I set a value `idTestTag` or sth so. This value will then be concatenated to AlertListsScreen, MyAlertItem, PalsAlertItem to have unique test tags for each alert that is displayed.  
So, if at some point we want to change the test tag that is concatenated, we don't have to go in the code and change it by hand one by one, but rather just change the `idTestTag`.

app/src/main/java/com/android/periodpals/ui/alert/AlertLists.kt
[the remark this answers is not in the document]
francelu: Same answer as before.

app/src/main/java/com/android/periodpals/ui/alert/AlertLists.kt
agonzalez-r: `String?` means that it is either a `String` or null, if it's null it means the alert doesn't exist in the repository and otherwise it will be a `String`

charliemangano: I believe the best way to handle that would be something like this:
```kotlin
if (alert.id == null) {
  // do smth
}
// use later on in the code like this:
alert.id!!
```
The bang bang operator tells the compiler that `alert.id` has been asserted as being non-null.

francelu: But if it doesn't exist in the repository, then I wouldn't be able to fetch it and get `alert.id = null`.

app/src/main/java/com/android/periodpals/ui/alert/AlertLists.kt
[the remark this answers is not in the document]
agonzalez-r: okay I see, but what is wrong with using the `alert.id` as "unique identifier" to add to the each test tag of the alerts (as you did)? I think it fulfils the purpose. And in that case, you pass just the alert as parameter and directly use the `alert.id`?

app/src/main/java/com/android/periodpals/ui/alert/AlertLists.kt
agonzalez-r: Yes, for the test tag it is indeed better to use the `alert.id`, leave it. What I meant was that we need to pass the `uid` in order to get the right Profile Picture (of the user) for each Alert. But I guess since that is not implemented yet you can leave it as it is

app/src/main/java/com/android/periodpals/ui/alert/AlertLists.kt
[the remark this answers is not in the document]
agonzalez-r: My comments was about already having the right parameters in each composables to implement all the needed features in the future.

app/src/main/java/com/android/periodpals/ui/alert/AlertLists.kt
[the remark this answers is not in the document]
charliemangano: I see, thanks a lot. It seems to have kept the behavior we want even with the changes you just made, well done!

app/src/main/java/com/android/periodpals/ui/alert/AlertLists.kt
agonzalez-r: if the id is null, then the alert won't even be displayed, we don't need to handle the case were the `alert.id` is null in this class. For the moment leave it as it is, we just need to remember to change it once we complete the implementation of the alert view model.

francelu: Did what @charliemangano suggested. Added TODO @agonzalez-r suggested.

app/src/main/java/com/android/periodpals/ui/alert/AlertLists.kt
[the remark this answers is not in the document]
francelu: I could.  
But even in that case, if you wanted to change it for example to uid, you would need to go through the entire code. Therefore I use an `idTestTag`.  

This part can be handled when the tests will be overwritten when the alert VM will be implemented, which will be done anyways.

app/src/main/java/com/android/periodpals/ui/alert/AlertLists.kt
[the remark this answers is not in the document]
francelu: Not in the scope of this task.

agonzalez-r (approved):
## Summary
This PR effectively fixes the issue with the layout out responsiveness to different screens! Very nice work on modifying the PR to mach the Figma, creating all the necessary components for this screen and advancing one step further into its completion! 
## Important 
#### Code Quality
I really like that you created composable functions in `AlertsLists.kt`. The code is properly modularised, consistent and clean. 
#### Functionality
Sadly I couldn't test the screen manually because of problems with my emulator, but from the code a your provided screenshots: LGTM !
## Testing
Tests are complete and thorough. They all pass! LGTM 
## Documentation
Missing documentation was fixed. LGTM
## Suggestions
- Pass the `alert` parameter instead of just `alert.id` on `AlertProfilePicture()`, `AlertTimeAndLocation()` ,`AlertProductAndUrgency()` and `AlertAcceptButtons()`, to facilitate implementation in the future. All of them will need access to other parameters of the Alert, and since you're already using `alert.id`, why not advance the code as much as possible and avoid updating documentation etc. 
- Need to be careful with `alert.id` type in future implementation

Issue #181, opened 2024-11-12
Add "Current Location" as a selectable option.


taghizadlaura:
### Description:
This is a subtask from #151. Add the "Current Location" option to the existing dropdown menu in the "Location" field on the CreateAlert screen. This will allow users to select "Current Location" from the predefined options in the dropdown.

### Done:
**Functional Requirements:**
- Add "Current Location" as an option in the existing dropdown menu in the Location field.
- The dropdown should expand to show the "Current Location" option when the user interacts with the Location field.

**Quality Standards:**
- Ensure smooth UI interactions for the dropdown menu with the new option.
- The dropdown should meet visual design standards to ensure a consistent user experience.
- Code quality and performance should meet the baseline project standards.

**Testing Requirements:**
- UI tests to verify that the dropdown menu appears and includes the "Current Location" option.
- Usability testing to confirm that the new option integrates well within the CreateAlert screen flow.

Issue #151, opened 2024-11-08
Add dropdown menu for CreateAlert screen location field with option "Current Location"


taghizadlaura:
### Description:
Add a dropdown menu to the "Location" field on the CreateAlert screen, allowing users to select "Current Location" as an option. This feature enhances the user experience by providing quick access to their current location, supporting faster alert creation. Note that the logic for fetching the current location will be implemented in a separate task; this task focuses solely on the UI of the dropdown button. This aligns with the sprint's goal of improving user interactions within the alert creation flow.

### Done:
- **Functional Requirements:**
  - Implement a dropdown menu in the Location field with at least one option: "Current Location."
  - When "Current Location" is selected, the field should display the label "Current Location" (location logic will be implemented separately).

- **Quality Standards:**
  - Ensure smooth UI responsiveness.
  - Handle errors gracefully if location access is unavailable or permissions are denied (although the logic for location fetching will be handled in a future task).
  - Code should meet project performance and readability standards.

- **Testing Requirements:**
  - Unit tests for dropdown menu behavior.
  - UI tests to ensure that the dropdown menu displays and updates correctly.

### Preconditions:
- Location permissions must be requested and granted prior to accessing "Current Location."
- This task assumes that location access logic and permission handling infrastructure will be implemented separately.

### Steps to Completion:
- [x] #152 
- [x] #181

### Points to Consider / Known Risks:
- Limited by the current design framework and potential delays in accessing location services.

PR #183, opened 2024-11-12
feat: add "Current Location" as a selectable option.

taghizadlaura:
# Add "Current Location" as a selectable option.

## Description
This PR introduces the UI for the "Use Current Location" feature on the Create Alert screen. Users can now see an option to set their location automatically using the current location, though the location logic is not yet implemented. This closes issue #181 (and so the issue #151).

## Changes
- Added a "Current Location" DropdownMenuItem to the Create Alert screen, allowing users to trigger the location setting.

#### Modified
- `CreateAlertScreen.kt`: Contains the "Current Location" option and a specific icon.

## Testing
- Since we didn't specifically test the dropdown items in any other field, I assumed that all dropdown-related tests would be handled later as part of the broader testing strategy.

### DIFF
--- a/app/src/main/java/com/android/periodpals/ui/alert/CreateAlert.kt
+++ b/app/src/main/java/com/android/periodpals/ui/alert/CreateAlert.kt
@@ -73,6 +73,8 @@ private const val SUBMISSION_BUTTON_TEXT = "Ask for Help"
 private const val MAX_NAME_LEN = 30
 private const val MAX_LOCATION_SUGGESTIONS = 3
 
+private const val CURRENT_LOCATION_TEXT = "Current Location"
+
 /**
  * Composable function for the CreateAlert screen.
  *
@@ -192,7 +194,7 @@ fun CreateAlertScreen(
               modifier = Modifier.wrapContentSize(),
           ) {
             DropdownMenuItem(
-                text = { Text("Current Location") },
+                text = { Text(CURRENT_LOCATION_TEXT) },
                 onClick = {
                   // Logic for fetching and setting current location will be implemented later
                   showDropdown = false // For now close dropdown on selection
--- a/app/src/main/java/com/android/periodpals/ui/alert/CreateAlert.kt
+++ b/app/src/main/java/com/android/periodpals/ui/alert/CreateAlert.kt
@@ -10,11 +10,14 @@ import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.layout.wrapContentHeight
 import androidx.compose.foundation.layout.wrapContentSize
 import androidx.compose.foundation.lazy.LazyColumn
+import androidx.compose.material.icons.Icons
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/android/periodpals/ui/alert/CreateAlert.kt
```diff
@@ -188,6 +191,16 @@ fun CreateAlertScreen(
               onDismissRequest = { showDropdown = false },
               modifier = Modifier.wrapContentSize(),
           ) {
+            DropdownMenuItem(
+                text = { Text("Current Location") },
```
charliemangano, line 195: Can you extract this to a constant?

app/src/main/java/com/android/periodpals/ui/alert/CreateAlert.kt
```diff
@@ -188,6 +191,16 @@ fun CreateAlertScreen(
...
+                onClick = {
+                  // Logic for fetching and setting current location will be implemented later
```
charliemangano, line 197: Nitpick but can you mark this comment as a TODO, it makes it easier to see what's left to do when planning tasks :)

charliemangano (approved):
LGTM, good job! Take a look at the comments for small nitpicks

francelu (approved):
## Summary
**LGTM**  
Minor suggestions/nitpicks: echoing what @charliemangano pointed out.  

Additionally, I noticed that this menu is a drop-down but it placed **up** the location field. It might be worth placing it below the location field for better UX, though I recognize this change may be out of scope for this PR. However, since you're working on location-related elements, it's something to consider in the future.


### Important (which includes)
- **Code Quality**: Good :)
- **Functionality**: Improved by this PR with the "Set current location"
- **Performance**: -
- **Security**: -

### Testing
Ensure adequate testing for all new changes and edge cases.

### Documentation
Verify if any updates to the documentation are necessary.

### Suggestions
Consider UX adjustments, particularly related to menu placement, in a future iteration.  
Also, it would be nice to have a consistent layout. (It is already good, but) Check the wiki page if you have time :)

### Steps before PR merge
- Explicitly address the suggestions and nitpicks raised.

taghizadlaura:
@francelu, you mentioned, 'I noticed that this menu is a drop-down but it’s placed above the location field. It might be worth placing it below the location field for better UX, though I recognize this change may be out of scope for this PR.' Do you want me to move it below all the options? Or should we wait for the Figma to be updated and discuss the UI later with the group?

francelu:
> @francelu, you mentioned, 'I noticed that this menu is a drop-down but it’s placed above the location field. It might be worth placing it below the location field for better UX, though I recognize this change may be out of scope for this PR.' Do you want me to move it below all the options? Or should we wait for the Figma to be updated and discuss the UI later with the group?

@taghizadlaura  
I think we will change the layout later on, when we will have updated the Figma. Same for padding, typography and colours.

Issue #164, opened 2024-11-09
Correctly moddify `Users` table


coaguila:
### Description:
Update `UserModel` and `UserViewModel.kt` so that it correctly modifies `Users` table in the Supabase database.

### Done:
Specific, measurable criteria to confirm the task is complete.
- Needs to correctly update a profile of a given `User`
- Resilient to Supabase API errors

### Steps to Completion:
Break down the task into actionable steps. Make sure those steps are separable so that you can easily create subtasks/subissues from these:
- [x] Implement function in `UserModel` & `UserModelSupabase` that checks wether `User` already exists
- [x] Implement function in `UserModel` & `UserModelSupabase` that updates a `User` if it exists
- [x] Implement funtion in `UserViewModel` that correctly updates `User` state

### Points to Consider / Known Risks:
Optional: Limited testing options

PR #186, opened 2024-11-13
feat/profile/upsert

coaguila:
# Add Upsert function to `UserModel` and `UserViewModel`

## Description
This PR introduces. It closes issue #164

## Changes
Adds `upsertUserProfile` to `UserModel` and implements it in `loadUserProfile` to `UserViewModel`.

#### Added
- none
#### Modified
- ``
## Dependencies Added
- dependency 1 for XXX
- ...

### DIFF
--- a/app/src/main/java/com/android/periodpals/model/user/UserModelSupabase.kt
+++ b/app/src/main/java/com/android/periodpals/model/user/UserModelSupabase.kt
@@ -57,4 +57,27 @@ class UserRepositorySupabase(private val supabase: SupabaseClient) : UserReposit
       onFailure(e)
     }
   }
+  override suspend fun updateUserProfile(
+      user: User,
+      onSuccess: () -> Unit,
+      onFailure: (Exception) -> Unit
+  ){
+    try{
+      withContext(Dispatchers.IO) {
+          supabase.postgrest[USERS].update(
+              {
+                  User::name setTo user.name
+                  User::imageUrl setTo user.imageUrl
+                  User::dob setTo user.dob
+                  User::description setTo user.description
+              }
+          ){
+              filter { User::name eq user }
+          }
+      }
+    } catch (e: Exception) {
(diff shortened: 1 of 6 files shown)


Harrish92 (approved):
## Summary
LGTM. Decide whether you want to keep `createUserProfile` for userModel.

## Suggestions
- Since upsert is similar to insert, `createUserProfile` may no longer be useful for the viewModel. Unless there is another usage or it is being used, remove it.

app/src/main/java/com/android/periodpals/model/user/UserModelSupabase.kt
```diff
@@ -57,4 +57,21 @@ class UserRepositorySupabase(private val supabase: SupabaseClient) : UserReposit
       onFailure(e)
     }
   }
+
+  override suspend fun upsertUserProfile(
```
charliemangano, line 61: I think having this function take a `UserDto` instead of a `User` like `createUserProfile` does is much better conceptually and separates duties more meaningfully, nice job!

Maybe think about doing the same in `createUserProfile`?

charliemangano (approved):
## Summary
Great job on adding this create-or-update alternative, it was much needed! Your code follows best practices and is well-documented :)  
## Important
#### Code Quality
Code is consistent with the rest of our codebase and follows best practices, nice job! I particularly appreciated that you extended the `User` class to add the `toUserDto` function.
#### Functionality
Code is correct and works well.
## Testing
Tests are updated to include checks for the new function, and are very complete. They all pass.
## Documentation
Docs is complete and comprehensive.
## Suggestions
Delete the now-obsolete `createUSerProfile`, or if we want to keep it just in case, complete the docs to say that it will throw an error if called with an already existing user, and change the parameter from `User` to `UserDto`
## Steps before PR approved
LGTM! Thanks for the quick work!

Issue #190, opened 2024-11-13
Fix sync problem with CreateProfile screen


charliemangano:
### Description:
We have noticed a problem with how we handle the creation of a user's profile, leading to having to click the "Save" button two times to properly navigate to the profile screen and display the data they've just entered. We believe it is caused by a syncing problem, either directly with our repository (Supabase), or in our code itself with side effects being handled improperly. This task aims at finding and fixing this problem.

### Done:
- **Functional Requirements:** 
  - creating the profile should work on the first try
  - user data should show up on Supabase database
- **Quality Standards:**
  - code should be consistent with the rest of the database
  - MVVM architecture
- **Testing Requirements:** 
  - will be handled in another task (end to end for M2)
  - code should still pass existing tests
  - should be > 80% line coverage

### Steps to Completion:
- [x] finding the problem
- [x] fixing the problem
- [ ] testing

charliemangano:
Same problem as the one described in #187.

PR #191, opened 2024-11-13
Fix/profile/sync-problems : Fix sync problem with CreateProfile screen

charliemangano:
# Fix sync problem with CreateProfile screen

## Description
This PR introduces a fix for issue #190: it solves the syncing problem that made a user have to click two times on the "Save" button on the CreateProfile screen.

The problem came from how we handle our side effects, as the callbacks of the view model execute, at least partly, only after we've checked in the CreateProfile screen that the saving had succeeded, for example:

```kotlin
// in `CreateProfile`
userViewModel.saveUser()              // this does not finish executing

if (userState.value == null) {...}    // before this happens
```
So our CreateProfile screen thought that the operation failed, when it had just not finished executing.

## Changes
Changed the signature of the functions in `UserViewModel` to have `onSuccess` and `onFailure` callbacks. I moved the code that was previously in the button's `onClick` into the callbacks of the calls to `saveUser` and `loadUser`. This ensures that this bit of code is executed **after** the rest of the `saveUser` or `loadUser` code, which solves the problem.
These callback all have simple `Log`s as default values. Docs were updated as well.
Updated the tests for the CreateProfile screen to implement the new callbacks properly.

#### Added
None
#### Modified
- `app/src/main/java/com/android/periodpals/model/user/UserViewModel.kt`: add callbacks to function signatures
- `app/src/main/java/com/android/periodpals/ui/profile/CreateProfile.kt`, `app/src/main/java/com/android/periodpals/ui/profile/ProfileScreen.kt`: update the calls to the view model functions to implement the callbacks properly
- `app/src/androidTest/java/com/android/periodpals/ui/profile/CreateProfileTest.kt`: update the tests to implement the callbacks properly

#### Removed
None

## Dependencies Added
None

## Testing
Existing tests were updated but still check for the same things. Tests all pass. None more written. This should help for the end-to-end tests.

# Edit: 14/11 8:22PM
Following a merge ([24910ae](https://github.com/PeriodPals/periodpals/pull/191/commits/24910aec9a2f79c66aab975926a23fc0de3e2483)), the function I was modifying (`onClick` of the "Save" `Button` of `CreateProfile`) got refactored into a different file (`ProfileComponents.kt`) that was now shared with `EditProfile`. This means that any modification I made to fix `CreateProfile`'s `onClick`, I now had to do in the `EditProfile` screen as well, and the associated tests. This PR therefore got extended after it had been approved. Having now modified everything I needed to, I'm requesting a re-review.

Updated list of modifications:
Everything from above, plus implementing changes asked in reviews, plus:
- `app/src/main/java/com/android/periodpals/ui/profile/CreateProfile.kt`: removed unused variable
- `app/src/main/java/com/android/periodpals/ui/profile/EditProfile.kt`: update call to `loadUser` to implement callbacks properly
- `app/src/main/java/com/android/periodpals/ui/components/ProfileComponents.kt`: updated now-refactored `ProfileSaveButton` with work done in previous, now overwritten commits (implement callbacks properly in call to `saveUser`)
- `app/src/androidTest/java/com/android/periodpals/ui/profile/EditProfileTest.kt`: updated to implement new callbacks properly 
- `app/src/test/java/com/android/periodpals/model/user/UserModelSupabaseTest.kt`: updated tests to run on main thread to comply with updated `UserModelSupabase`(fix to unit tests not passing after [dc06a22](https://github.com/PeriodPals/periodpals/pull/191/commits/dc06a22537de25d6a06718e119add4c49ce3866e))

### DIFF
--- a/app/src/androidTest/java/com/android/periodpals/ui/profile/CreateProfileTest.kt
+++ b/app/src/androidTest/java/com/android/periodpals/ui/profile/CreateProfileTest.kt
@@ -119,7 +119,7 @@ class CreateProfileTest {
         .performTextInput(description)
     composeTestRule.onNodeWithTag(ProfileScreens.SAVE_BUTTON).performClick()
 
-    verify(userViewModel, never()).saveUser(any())
+    verify(userViewModel, never()).saveUser(any(), any(), any())
 
     verify(navigationActions, never()).navigateTo(any<TopLevelDestination>())
     verify(navigationActions, never()).navigateTo(any<String>())
@@ -136,7 +136,7 @@ class CreateProfileTest {
         .performTextInput(description)
     composeTestRule.onNodeWithTag(ProfileScreens.SAVE_BUTTON).performClick()
 
-    verify(userViewModel, never()).saveUser(any())
+    verify(userViewModel, never()).saveUser(any(), any(), any())
 
     verify(navigationActions, never()).navigateTo(any<TopLevelDestination>())
     verify(navigationActions, never()).navigateTo(any<String>())
@@ -151,7 +151,7 @@ class CreateProfileTest {
     composeTestRule.onNodeWithTag(ProfileScreens.NAME_INPUT_FIELD).performTextInput(name)
     composeTestRule.onNodeWithTag(ProfileScreens.SAVE_BUTTON).performClick()
 
-    verify(userViewModel, never()).saveUser(any())
+    verify(userViewModel, never()).saveUser(any(), any(), any())
 
     verify(navigationActions, never()).navigateTo(any<TopLevelDestination>())
     verify(navigationActions, never()).navigateTo(any<String>())
@@ -160,6 +160,10 @@ class CreateProfileTest {
   @Test
   fun createValidProfileVMFailure() {
     `when`(userViewModel.user).thenReturn(mutableStateOf(null))
+    `when`(userViewModel.saveUser(any(), any(), any())).thenAnswer {
+      val onFailure = it.arguments[2] as (Exception) -> Unit
+      onFailure(Exception("Error saving user"))
+    }
     composeTestRule.setContent { CreateProfileScreen(userViewModel, navigationActions) }
 
     composeTestRule.onNodeWithTag(ProfileScreens.DOB_INPUT_FIELD).performTextInput(dob)
@@ -169,7 +173,7 @@ class CreateProfileTest {
         .performTextInput(description)
     composeTestRule.onNodeWithTag(ProfileScreens.SAVE_BUTTON).performClick()
 
-    verify(userViewModel).saveUser(any())
+    verify(userViewModel).saveUser(any(), any(), any())
 
     verify(navigationActions, never()).navigateTo(Screen.PROFILE)
   }
@@ -177,6 +181,10 @@ class CreateProfileTest {
   @Test
   fun createValidProfileVMSuccess() {
     `when`(userViewModel.user).thenReturn(userState)
+    `when`(userViewModel.saveUser(any(), any(), any())).thenAnswer {
+      val onSuccess = it.arguments[1] as () -> Unit
+      onSuccess()
+    }
     composeTestRule.setContent { CreateProfileScreen(userViewModel, navigationActions) }
 
     composeTestRule.onNodeWithTag(ProfileScreens.DOB_INPUT_FIELD).performTextInput(dob)
@@ -186,7 +194,7 @@ class CreateProfileTest {
         .performTextInput(description)
     composeTestRule.onNodeWithTag(ProfileScreens.SAVE_BUTTON).performClick()
 
-    verify(userViewModel).saveUser(any())
+    verify(userViewModel).saveUser(any(), any(), any())
 
     verify(navigationActions).navigateTo(Screen.PROFILE)
   }
--- a/app/src/main/java/com/android/periodpals/model/user/UserModelSupabase.kt
+++ b/app/src/main/java/com/android/periodpals/model/user/UserModelSupabase.kt
@@ -22,7 +22,7 @@ class UserRepositorySupabase(private val supabase: SupabaseClient) : UserReposit
   ) {
     try {
       val result =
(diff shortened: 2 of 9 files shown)


app/src/androidTest/java/com/android/periodpals/ui/profile/CreateProfileTest.kt
```diff
@@ -169,14 +173,18 @@ class CreateProfileTest {
...
+    verify(userViewModel).saveUser(any(), any(), any())
 
     verify(navigationActions, never()).navigateTo(Screen.PROFILE)
   }
 
   @Test
   fun createValidProfileVMSuccess() {
     `when`(userViewModel.user).thenReturn(userState)
+    `when`(userViewModel.saveUser(any(), any(), any())).thenAnswer {
+      val onSuccess = it.arguments[1] as () -> Unit
```
agonzalez-r, line 230: can you please explain what the `it.arguments[1]` is?

charliemangano: Arguments of the call to `UserViewModel.saveUser` are stored in the `arguments` list. This fetches the argument at index 1, which is the `onSuccess` callback that I'm calling in the next line. Does that answer your questions?

Harrish92 (changes requested):
## Summary
Good job on the fix. Just need some required adjustments and clarifications.

## Important
#### Code Quality
- Code is well structured. Though some parts need clarifications.
## Suggestions
- Add some comments in the code : e.g. for `Handler` and `Looper`
## Steps before PR approved
- Remove unnecessary `userState` variable in `CreateProfile`.
-  Do what is being asked in the review comments or give an explanation.

app/src/main/java/com/android/periodpals/ui/profile/CreateProfile.kt
```diff
@@ -172,16 +174,21 @@ private fun attemptSaveUserData(
...
-
-  Log.d(LOG_TAG, LOG_SUCCESS)
-  Toast.makeText(context, TOAST_SUCCESS, Toast.LENGTH_SHORT).show()
-  navigationActions.navigateTo(Screen.PROFILE)
+  userViewModel.saveUser(
+      user = newUser,
+      onSuccess = {
+        Handler(Looper.getMainLooper()).post {
+          Toast.makeText(context, TOAST_SUCCESS, Toast.LENGTH_SHORT).show()
+        }
```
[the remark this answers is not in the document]
charliemangano: Toast can't be called from a thread that's not the main UI thread. Since this callback is executed in the background (so not the main UI thread), we need to "send" it's run to the main thread. That's what this piece of code does.

We might not need it later on if we decide to inform the users using something else than toasts.

app/src/main/java/com/android/periodpals/ui/profile/ProfileScreen.kt
```diff
@@ -70,12 +72,16 @@ fun ProfileScreen(userViewModel: UserViewModel, navigationActions: NavigationAct
       0 // TODO: placeholder to be replaced when we integrate it to the User data class
 
   Log.d(TAG, "Loading user data")
-  userViewModel.loadUser()
+  userViewModel.loadUser(
+      onSuccess = {},
+      onFailure = { e ->
```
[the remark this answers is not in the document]
charliemangano: You're right, it should use the default `onSuccess` defined in the `UserViewModel`.

Harrish92 (approved):
LGTM

app/src/androidTest/java/com/android/periodpals/ui/profile/CreateProfileTest.kt
agonzalez-r: Yes thank you!

agonzalez-r (approved):
The bug is gone. Code works fine now, good work!
LGTM

app/src/main/java/com/android/periodpals/model/user/UserViewModel.kt
```diff
@@ -39,18 +52,28 @@ class UserViewModel(private val userRepository: UserRepositorySupabase) : ViewMo
    * Saves the user profile.
    *
    * @param user The user profile to be saved.
+   * @param onSuccess Callback function to be called when the user profile is successfully saved.
+   * @param onFailure Callback function to be called when there is an error saving the user profile.
    */
-  fun saveUser(user: User) {
+  fun saveUser(
+      user: User,
+      onSuccess: () -> Unit = { Log.d(TAG, "saveUser success callback") },
```
coaguila, line 60: Great idea of using default values, we should do a future issue that does this for select issues

app/src/main/java/com/android/periodpals/model/user/UserModelSupabase.kt
coaguila, line 1: It think that `withContext` changes the `Dispatcher` used, perhaps we should reevaluate which `Dispatchers` are used when but in a future issue

app/src/main/java/com/android/periodpals/model/user/UserViewModel.kt
coaguila, line 1: Great idea to add the callbacks to the viewmodel!

coaguila (approved):
## Summary
For the most part, seems correct, perhaps we will have to reevaluate how we use `Dispatchers` within our codebase. Thanks for the fix!
## Important (which includes)
#### Code Quality
As always easy to read and seems to follow how we have set up our code. Even the new addition of callback to `UserViewModel`, while new to other viewmodels, it is consistent with how we have coded other callbacks.
#### Functionality
The addition to the callbacks fixes the sign up issue and avoids loading issues we had. Thanks again for adding this.
#### Performance (if applicable)
Perhaps we need to reevaluate what dispatchers we use to optimize io operations in a future pr.
## Testing
Correctly updated the m and vm from `User`
## Documentation
## Suggestions
Further research into correctly using `Dispatchers` within our model codebase.

charliemangano:
Merging this but we will definitely create a task to see what to do about the dispatchers @coaguila. Thanks for the last minute review!

Issue #195, opened 2024-11-14
Fix location fetching issue


taghizadlaura:
### Description:
Resolve issues related to the location fetching feature within the `LocationModelNominatim` class. The current implementation does not consistently retrieve location suggestions. This fix aims to improve the reliability of location data retrieval, ensuring accurate and complete results for users searching for location-based data. The task aligns with the sprint goal of enhancing user experience in map-based features.

### Done:
The task is considered complete when the following criteria are met:

- **Functional Requirements:**  
  - Location suggestions appear consistently and accurately for various input queries.
  - The system gracefully handles errors or empty responses, providing informative feedback.
  - The URL used for fetching data is verified, correctly formatted, and logged for debugging.

- **Quality Standards:**  
  - Ensure code quality and readability, adhering to existing code style guidelines.
  - Verify performance remains within acceptable limits when fetching location data.
  - Robust error handling is implemented for network issues or malformed responses.

- **Testing Requirements:**  
  - Unit tests cover successful data fetches and error-handling scenarios.
  - Code coverage meets or exceeds the project’s standards for this functionality.
  
### Preconditions:
- Dependencies: Ensure the OkHttp library is correctly configured in the project.
- Access to Nominatim API or a valid testing setup for external API requests.

### Steps to Completion:
- [ ] Add a User-Agent

PR #196, opened 2024-11-14
Fix location fetching issue

taghizadlaura:
# Fix location fetching issue : add a User-Agent

## Description
This PR introduces a User-Agent header to the HTTP request, which resolves the issue where location data was not being fetched correctly from the Nominatim API. By including a User-Agent, our requests are now compliant with the API’s requirements, enabling proper responses. This change addresses issue #195

## Changes
- Added a User-Agent header to the HTTP request in `LocationModelNominatim`.

#### Modified
- `LocationModelNominatim.kt`: Updated the `search` function to include a User-Agent in the request header.

## Dependencies Added
_No new dependencies added._

## Testing
- Verified that the location data is now fetched successfully with the User-Agent header.
- Confirmed that no errors occur with the modified request, and location suggestions appear as expected.


charliemangano (approved):
Good work finding the root of the issue!
LGTM, all tests still pass. Ran `MainActivity` on my end and it seems to work now.

Issue #231, opened 2024-11-21
Fix and Optimize Alert View Model


coaguila:
## Description
Currently some functions within `AlertViewModel` return values and don't follow the callback structure we have been using in other parts of the code base. Also `getPalAlerts` returns a value and doesnt store it within a state. Also there might be better use of states.

## Done
Specific, measurable criteria to confirm the task is complete.
- Refactor all functions that return value to use callbacks
- Have clear documentation and callbacks
- Test potential new states

## Steps to Completion
Break down the task into actionable steps. Make sure those steps are separable so that you can easily create subtasks/subissues from these:
- [x] Refactor to callbacks
- [x] Look into changing states for `alert` in the view model
- [x] Add a derieved state to replace the `getPalAlerts` function
- [x] Update tests

PR #241, opened 2024-11-25
Fix Alert View Model

coaguila:
# Fix Alert View Model

## Description
This PR fixes some issues the `AlertViewModel`, notably the lack of consistency with previous code that uses callbacks, functions that return values, unoptimized use of `palsLists`. To update the `alerts` state, call `loadAlerts`, make sure to regualarly call it to have updated values. All functions that update the alerts list call `loadAlerts`. It closes issue #231 

## Changes
Removed `getAllAlerts` and replaced it with `fetchAlerts` its purpose is to update the `alerts` property, it is called on all the functions that modify the list. This function is left public to allow UI to refresh or even implement a refresh button feature.
#### Modified
- `app/src/main/java/com/android/periodpals/model/alert/AlertViewModel.kt`
- `app/src/test/java/com/android/periodpals/model/alert/AlertViewModelTest.kt`

## Testing
Tested that model functions are called, implemented more check within tests to insure code is working correctly. Removed test for deleted funtions and added functionality to mocked functions.

### DIFF
--- a/app/src/main/java/com/android/periodpals/model/alert/AlertViewModel.kt
+++ b/app/src/main/java/com/android/periodpals/model/alert/AlertViewModel.kt
@@ -178,8 +178,18 @@ class AlertViewModel(
   }
 
   /**
-   * Sets the filter for the filter alert list.
-   *
+   * Sets the filter for the `filterAlerts` list. Some filter examples could be:
+   * ```
+   * alertViewModel.setFilter(filter = { it.urgency == Urgency.HIGH } )
+   * // or
+   * alertViewModel.setFilter(filter = { it.product == Product.TAMPON )
+   * ```
+   * You could also make more complex filters like:
+   * ```
+   * alertViewModel.setFilter(
+   *    filter = { it.urgency == Urgency.HIGH && it.product == Product.TAMPON}
+   * )
+   * ```
    * @param filter Filter to be set on the alert list
    */
   fun setFilter(filter: (Alert) -> Boolean) {
--- a/app/src/main/java/com/android/periodpals/model/alert/AlertViewModel.kt
+++ b/app/src/main/java/com/android/periodpals/model/alert/AlertViewModel.kt
@@ -14,14 +14,21 @@ private const val TAG = "AlertViewModel"
  * ViewModel for managing alert data.
  *
  * @property alertModelSupabase The repository used for loading and saving alerts.
+ * @property userId the id linked to the current user
  * @property _alerts Mutable state holding the list of alerts.
  * @property alerts Public state exposing the list of alerts.
+ * @property _myAlerts Mutable state holding the list of current users alerts.
+ * @property myAlerts Public state exposing the list of current users alerts.4
+ * @property _palAlerts Mutable state holding the list of other users alerts.
+ * @property palAlerts Public state exposing the list of other users alerts.
+ * @property alertFilter Mutable state holding a filter for `filterAlerts`
+ * @property _filterAlerts Mutable state holding the list of alerts filtered by `alertFilter`
+ * @property filterAlerts Public state exposing the list of alerts filtered y `alertFilter`
  */
 class AlertViewModel(
     private val alertModelSupabase: AlertModelSupabase,
     private val userId: String
 ) : ViewModel() {
-  // remove this?
   private var _alerts = mutableStateOf<List<Alert>>(listOf())
   val alerts: State<List<Alert>> = _alerts
 
@@ -117,7 +124,6 @@ class AlertViewModel(
   ) {
     viewModelScope.launch {
       alertModelSupabase.getAlertsFilteredBy(
-          // ideally the uid would not be passed as argument and instead we could get uid by
           cond = { eq("uid", uid) },
           onSuccess = { alerts ->
             Log.d(TAG, "getAlertsByUser: Success")
@@ -134,6 +140,7 @@ class AlertViewModel(
    * Updates an existing alert.
    *
    * @param alert The alert with updated parameters.
+   * @param onSuccess Callback function to be called on success.
    * @param onFailure Callback function to be called on failure.
    */
   fun updateAlert(alert: Alert, onSuccess: () -> Unit, onFailure: (Exception) -> Unit) {
--- a/app/src/main/java/com/android/periodpals/model/alert/AlertViewModel.kt
+++ b/app/src/main/java/com/android/periodpals/model/alert/AlertViewModel.kt
@@ -158,12 +158,14 @@ class AlertViewModel(
    * // or
    * alertViewModel.setFilter(filter = { it.product == Product.TAMPON )
    * ```
+   *
    * You could also make more complex filters like:
    * ```
    * alertViewModel.setFilter(
    *    filter = { it.urgency == Urgency.HIGH && it.product == Product.TAMPON}
    * )
    * ```
+   *
    * @param filter Filter to be set on the alert list
    */
   fun setFilter(filter: (Alert) -> Boolean) {
--- a/app/src/main/java/com/android/periodpals/model/alert/AlertViewModel.kt
+++ b/app/src/main/java/com/android/periodpals/model/alert/AlertViewModel.kt
@@ -2,6 +2,7 @@ package com.android.periodpals.model.alert
 
 import android.util.Log
 import androidx.compose.runtime.State
+import androidx.compose.runtime.derivedStateOf
 import androidx.compose.runtime.mutableStateOf
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/android/periodpals/model/alert/AlertViewModel.kt
```diff
@@ -16,127 +17,168 @@ private const val TAG = "AlertViewModel"
  * @property _alerts Mutable state holding the list of alerts.
  * @property alerts Public state exposing the list of alerts.
  */
-class AlertViewModel(private val alertModelSupabase: AlertModelSupabase) : ViewModel() {
+class AlertViewModel(
+    private val alertModelSupabase: AlertModelSupabase,
+    private val userId: String
```
agonzalez-r, line 22: Please add documentation for the `userId` parameter

app/src/main/java/com/android/periodpals/model/alert/AlertViewModel.kt
```diff
@@ -16,127 +17,168 @@ private const val TAG = "AlertViewModel"
...
   /**
    * Retrieves all alerts.
    *
+   * @param onSuccess Callback function to be called on success, use the states to get results
+   * @param onFailure Callback function to be called on failure
    * @return The list of all alerts.
    */
-  fun getAllAlerts(): List<Alert>? {
-    var alertsList: List<Alert>? = null
+  fun fetchAlerts(onSuccess: () -> Unit = {}, onFailure: (Exception) -> Unit = {}) {
```
agonzalez-r, line 90: Maybe we can make this function private as Android Studio suggests

app/src/main/java/com/android/periodpals/model/alert/AlertViewModel.kt
```diff
@@ -16,127 +17,168 @@ private const val TAG = "AlertViewModel"
...
+  private var _alerts = mutableStateOf<List<Alert>>(listOf())
+  val alerts: State<List<Alert>> = _alerts
+
+  private var _myAlerts = derivedStateOf<List<Alert>> { _alerts.value.filter { it.uid == userId } }
+  val myAlerts: State<List<Alert>> = _myAlerts
+
+  private var _palAlerts = derivedStateOf<List<Alert>> { _alerts.value.filter { it.uid != userId } }
+  val palAlerts: State<List<Alert>> = _palAlerts
+
+  private var alertFilter = mutableStateOf<(Alert) -> Boolean>({ false })
```
agonzalez-r, line 34: Also add documentation for this variable, to make teh code more readable and understandable to others

app/src/main/java/com/android/periodpals/model/alert/AlertViewModel.kt
```diff
@@ -16,127 +17,168 @@ private const val TAG = "AlertViewModel"
...
-  fun getAlertsByUser(uid: String): List<Alert>? {
-    var alertsList: List<Alert>? = null
+  fun getAlertsByUser(
+      uid: String,
+      onSuccess: (List<Alert>) -> Unit,
+      onFailure: (Exception) -> Unit
+  ) {
     viewModelScope.launch {
       alertModelSupabase.getAlertsFilteredBy(
           // ideally the uid would not be passed as argument and instead we could get uid by
```
agonzalez-r, line 120: Remove useless comment

app/src/main/java/com/android/periodpals/model/alert/AlertViewModel.kt
```diff
@@ -16,127 +17,168 @@ private const val TAG = "AlertViewModel"
...
+          })
     }
-    return alertsList
   }
 
   /**
    * Updates an existing alert.
    *
    * @param alert The alert with updated parameters.
+   * @param onFailure Callback function to be called on failure.
```
agonzalez-r, line 137: I think you forgot documentation for the `onSuccess` parameter

app/src/main/java/com/android/periodpals/model/alert/AlertViewModel.kt
```diff
@@ -16,127 +17,168 @@ private const val TAG = "AlertViewModel"
...
+          })
     }
   }
+
+  /**
+   * Sets the filter for the filter alert list.
+   *
+   * @param filter Filter to be set on the alert list
+   */
+  fun setFilter(filter: (Alert) -> Boolean) {
```
agonzalez-r, line 181: Nice!

app/src/test/java/com/android/periodpals/model/alert/AlertViewModelTest.kt
```diff
@@ -124,8 +109,10 @@ class AlertViewModelTest {
         .`when`(alertModelSupabase)
         .addAlert(any<Alert>(), any<() -> Unit>(), any<(Exception) -> Unit>())
 
-    viewModel.createAlert(alert)
-    assert(viewModel.alerts.value!!.isEmpty())
+    viewModel.createAlert(alerts[0], { fail("Should not call `onFailure`") }, {})
```
charliemangano, line 112: I believe the failure message here should be "Should not call `onSuccess``"

app/src/main/java/com/android/periodpals/model/alert/AlertViewModel.kt
coaguila: Strongly disagree this allows the ui to update the list by themself, while it is called when doing changes to list, it is important for the UI people to be allowed to update the list without doing a change to the list.

agonzalez-r: Okay, I see ! should the function be tested in that case?

app/src/test/java/com/android/periodpals/model/alert/AlertViewModelTest.kt
```diff
@@ -161,33 +151,28 @@ class AlertViewModelTest {
...
 
-    doAnswer { it.getArgument<(List<Alert>) -> Unit>(0)(listOf(alert)) }
-        .`when`(alertModelSupabase)
-        .getAllAlerts(any<(List<Alert>) -> Unit>(), any<(Exception) -> Unit>())
+    viewModel.deleteAlert(id[0], {}, { fail("Should not `onFailure`") })
 
-    viewModel.createAlert(alert)
-    viewModel.deleteAlert(alertNullID)
+    verify(alertModelSupabase)
+        .deleteAlertById(any<String>(), any<() -> Unit>(), any<(Exception) -> Unit>())
```
charliemangano, line 173: This should check for the correctness of the arguments:
```kotlin
verify(alertModelSupabase).deleteAlertsById(eq(id[0]), any<() -> Unit>(), any<(Exception) -> Unit>())
```

app/src/test/java/com/android/periodpals/model/alert/AlertViewModelTest.kt
```diff
@@ -104,17 +90,16 @@ class AlertViewModelTest {
...
-          onSuccess(listOf(alert)) // Return a list with our mock alert
-          null
-        }
+    doAnswer { it.getArgument<(List<Alert>) -> Unit>(0)(listOf(alerts[0])) }
         .`when`(alertModelSupabase)
-        .getAllAlerts(any(), any())
+        .getAllAlerts(any<(List<Alert>) -> Unit>(), any<(Exception) -> Unit>())
 
-    viewModel.createAlert(alert)
+    viewModel.createAlert(alerts[0], {}, { fail("Should not call `onFailure`") })
```
agonzalez-r, line 97: You can create constants for the error messages that are commonly reused, such as "Should not call `onFailure`"

app/src/main/java/com/android/periodpals/model/alert/AlertViewModel.kt
charliemangano: I'll be leaving a comment about this in my own review but I agree with @agonzalez-r that it should be made private since in principle it should be called systematically inside every function of the VM.

Private or not, it should still be tested in some capacity.

app/src/main/java/com/android/periodpals/model/alert/AlertViewModel.kt
```diff
@@ -13,130 +14,178 @@ private const val TAG = "AlertViewModel"
  * ViewModel for managing alert data.
  *
  * @property alertModelSupabase The repository used for loading and saving alerts.
+ * @property userId the id linked to the current user
  * @property _alerts Mutable state holding the list of alerts.
  * @property alerts Public state exposing the list of alerts.
+ * @property _myAlerts Mutable state holding the list of current users alerts.
+ * @property myAlerts Public state exposing the list of current users alerts.4
```
agonzalez-r, line 21: nitpick: you left a 4 there lol

agonzalez-r (approved):
## Summary
This PR effectively addresses the issues found on the previous version of `AlertViewModel` by using callbacks in functions, ensuring consistency. It introduces new properties to the view model that made its usability much more functional (`myAlerts`, `palAlerts`, uid passed as parameter...). 

Thank you for your hard work @coaguila !! the effort you put into this is very clear. 
## Important 
#### Code Quality
- Readability: The code is well-structured and readable after documentation was added. 
- Consistency: The coding style is consistent with the rest of the codebase.
#### Functionality
- Correctness: The code is well implemented. 
- Edge Cases: This could probably be improved, notably by adding null and empty handling in `_filterAlerts` to ensure robustness. For example: 
```
private var _filterAlerts = derivedStateOf { 
    _alerts.value?.filter { alertFilter.value(it) } ?: emptyList()
}
```
This ensures that if _alerts.value is null, it returns an empty list instead of causing a potential null pointer exception.
#### Performance
- The use of derivedStateOf for filtering alerts is efficient and should improve performance by reducing unnecessary recomputations.
## Testing
- Tests are very good, clever implementation. They all pass locally 
- Missing tests for `myAlerts` and `palAlerts` lists, although the code seems correct
- Missing tests for `fetchAlerts()` since it will be used outside the view model class, as you suggested in the comments 
## Documentation
- Missing documentation was added, everything seems complete
## Suggestions
- Implement missing tests.
- Also, now that you added a filtering function, do you think the function `getAlertsByUser()` is still necessary?
## Steps before PR approved
I'm approving directly because I do not think the missing tests are a big concern and I now that this PR is kind of urgent for other tasks to be done. The coverage is already correct and I think the implemented tests are very complete.

charliemangano (changes requested):
## Summary
Exactly what we needed done, nice job! You're successfully resolving the issues we had with the previous verion of the VM with the callbacks. Thanks for all the work!

However, I do have some comments.

## Important
#### Code Quality
Code is consistent with our codebase and very readable, well done. As pointed out by @agonzalez-r, please clean up the useless comments.

#### Functionality
As said in the comments, it seems to me that `fetchAlerts` should be `private` **and** called systematically in every VM method, otherwise it makes no sense structurally and doesn't respect our MVVM architecture as much as it could. It should not be left up to the view to update our VM attributes, this behaviour should be entirely controlled by the VM itself.

To be honest, I'm not entirely sure whether `fetchAlerts` and the vm attributes are even useful, since we are handling everything within the callbacks now. When I updated the userVM, I left them since I wasn't quite sure of my implementation but I'm becoming increasingly convinced that we don't need them.

Other than that, great job on implementing the fixes we needed in this VM!

## Testing
Very thorough tests, good job! 

Some remark however: 
 - "Positive" `verify`s (as in, calls to `verify` that check that a certain function has been called, as opposed to "negative" `verify`s, where we pass `never()` as an argument) should check for the correctness of the arguments as far as possible. I left an example of this in the comments. On the other hand, "negative" `verify`s should generally check for any arguments passed.
- The method `fetchAlerts` should be tested, as pointed out by @agonzalez-r.

## Documentation
Take care to update the docs where needed. I pointed it out in the comments where I saw it so you can refer to that.

## Suggestions
Please consider what I highlighted in the "Functionality" and "Testing" sections

## Steps before PR approved
See comments, plus what was mentionned above.

app/src/main/java/com/android/periodpals/model/alert/AlertViewModel.kt
```diff
@@ -16,127 +17,168 @@ private const val TAG = "AlertViewModel"
...
    * Retrieves an alert by its ID.
    *
    * @param idAlert The ID of the alert to be retrieved.
+   * @param onSuccess Callback function called upon successful run of function
+   * @param onFailure Callback function called upon a failed run of function
    * @return The alert if found, null otherwise.
    */
-  fun getAlert(idAlert: String): Alert? {
-    var alert: Alert? = null
+  fun getAlert(idAlert: String, onSuccess: (Alert) -> Unit, onFailure: (Exception) -> Unit) {
```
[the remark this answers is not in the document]
coaguila: As I responded to Ale's previous comment on the subject: "Strongly disagree this allows the ui to update the list by themself, while it is called when doing changes to list, it is important for the UI people to be allowed to update the list without doing a change to the list."

app/src/main/java/com/android/periodpals/model/alert/AlertViewModel.kt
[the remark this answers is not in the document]
coaguila: As I responded to her message before: "Strongly disagree this allows the ui to update the list by themself, while it is called when doing changes to list, it is important for the UI people to be allowed to update the list without doing a change to the list."

app/src/main/java/com/android/periodpals/model/alert/AlertViewModel.kt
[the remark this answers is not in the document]
coaguila: I could envision different filters that UI such as filtering by different properties such as product type and urgency, this allows modularity for a future search bar

Issue #243, opened 2024-11-26
Use Location data class everywhere in the code and delete GPSLocation


lazarinibruno:
Currently, the code has two location data classes: `Location` and `GPSLocation`. The only difference between both is that the former has a `name: String` attribute that contains the name (i.e. address) associated to the coordinates. 

Since we work with these two, both need methods to convert from one another. This is cumbersome and makes the code unclear. An easy solution is to set the `name` of the Location to "Current location" whenever the data class is used for the GPS.

PR #244, opened 2024-11-26
feat: delete GPSLocation and replace it by Location

lazarinibruno:
# Delete GPSLocation from the code and replace it by Location 

## Description
This PR deletes the `GPSLocation` data class and replaces all its usages by the `Location` data class. It closes issue #243.

We decided to do so because the only difference between both classes was that `Location` had an extra attribute `name` used to store the address associated with the location. The `LocationModelNominatim` used `Location` while the `GPSService` used `GPSLocation`. This made it very cumbersome to work with locations in the app, as we had to constantly convert from one to the other. It is highly probable that this would eventually lead to bugs.

In addition to this, I also added the serialization methods for the `Location` data class in preparation for the Alert ViewModel integration.  

#### Added
- file 1
- file 2
- ...
## Dependencies Added
- dependency 1 for XXX
- ...


app/src/test/java/com/android/periodpals/model/location/LocationDataClassTest.kt
```diff
@@ -0,0 +1,125 @@
...
+   */
+  private fun isSerializationCorrect(locations: List<Location>): Boolean {
+    for (loc in locations) {
+      val serialized = loc.toString()
+      val deserialized = Location.fromString(serialized)
+
+      if (deserialized.latitude != loc.latitude ||
+          deserialized.longitude != loc.longitude ||
+          deserialized.name != loc.name)
+          return false
```
agonzalez-r, line 121: nitpick: can't you just return 
```
!(deserialized.latitude != loc.latitude || deserialized.longitude != loc.longitude || deserialized.name != loc.name)
```
Instead of using `if(...)`

app/src/main/java/com/android/periodpals/model/location/Location.kt
```diff
@@ -1,7 +1,66 @@
...
+    val escapedName = name.replace(",", COMMA_URL_ENCODING)
+    return "$latitude,$longitude,$escapedName"
+  }
+
+  /**
+   * Transform this location into a [GeoPoint].
+   *
+   * @return The respective [GeoPoint].
+   */
+  fun toGeoPoint() = GeoPoint(latitude, longitude)
```
agonzalez-r, line 35: I'm kind of confused by the code, we don't need a function `fromGeoPoint()` (to `Location`)?

agonzalez-r: Is this what you mean in the description when you say: 

> we need to implement a "backwards search" method in the LocationModel

?

agonzalez-r (approved):
## Summary
Thank you for adding serialization/deserialization for the Location! it is indeed very important to have this for the creation of Alerts.
## Important 
#### Code Quality
- LGTM
#### Functionality
- little nitpick in the comments, overall good!
## Testing
- Very thorough tests! they all pass locally as well 
## Documentation
- Documentation is complete. 
## Suggestions
None, just some questions in the comments

app/src/main/java/com/android/periodpals/model/location/Location.kt
```diff
@@ -1,7 +1,66 @@
 package com.android.periodpals.model.location
 
-data class Location(
-    val latitude: Double,
-    val longitude: Double,
-    val name: String,
-)
+import org.osmdroid.util.GeoPoint
+
+private const val PARTS_ERROR_MESSAGE = "Invalid format. Expected 'lat,long'."
```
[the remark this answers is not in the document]
lazarinibruno: Good catch!

app/src/main/java/com/android/periodpals/model/location/Location.kt
lazarinibruno: `toGeopoint()` is a function from OSMDroid to display the markers in the map. We don't need a `fromGeopoint()` function since the alerts' location come from the database. Its not like we "have" the location data in the map in the form of GeoPoints and then convert them to `Location` objects using `fromGeopoint()`.

What I meant by the backwards search is in the following line of code in the `GPSServiceImpl.kt` file:

```kotlin
// TODO change CURRENT_LOCATION_NAME to actual location based on the coordinates
// CURRENT_LOCATION_NAME = "Current location"
_location.value = Location(lat, long, Location.CURRENT_LOCATION_NAME)
```

Instead of writing `Location.CURRENT_LOCATION_NAME` do something like: 

```kotlin
// addressFromCoordinates finds the closes address to the specified lat and long coordinates
_location.value = Location(lat, long, addressFromCoordinates(lat, long))
```

Let me know if its clearer now.

app/src/test/java/com/android/periodpals/model/location/LocationDataClassTest.kt
agonzalez-r: what I proposed is equivalent to return : 
```
deserialized.latitude == loc.latitude && deserialized.longitude == loc.longitude && deserialized.name == loc.name
```
(becase !( a!=b || c!=d ) is equivalent to( a==b && c ==d )). Use whatever form is more intuitive for you.
But all the conditions would have to be true to return true.

app/src/main/java/com/android/periodpals/model/location/Location.kt
agonzalez-r: Oh I see, I got confused sorry!

app/src/main/java/com/android/periodpals/model/location/Location.kt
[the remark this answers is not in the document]
francelu: This isn't fixed yet, is it ?

app/src/main/java/com/android/periodpals/model/location/Location.kt
```diff
@@ -1,7 +1,66 @@
...
+private const val STRING_DELIMITER = ","
+private const val COMMA_URL_ENCODING = "%2C"
+
+/**
+ * Represents a geographic location associated to a name.
+ *
+ * @property latitude the latitude of the location
+ * @property longitude the longitude of the location
+ */
+data class Location(val latitude: Double, val longitude: Double, val name: String) {
```
francelu, line 16: Missing documentation for `name`.

app/src/main/java/com/android/periodpals/model/location/Location.kt
```diff
@@ -1,7 +1,66 @@
...
+
+  /**
+   * Converts the GPSLocation object to a String in "latitude,longitude, name" format. Escapes
+   * commas in the [name].
+   *
+   * @return A String representation of the GPSLocation object.
+   */
+  override fun toString(): String {
+    // Replace comma with its URL encoding counterpart
```
francelu, line 25: **Nitpick**: Remove this commented line as the function's documentation already should explain what it does.

app/src/main/java/com/android/periodpals/model/location/Location.kt
```diff
@@ -1,7 +1,66 @@
...
+      val long = parts[1].toDoubleOrNull()
+      // Replace URL-encoded comma with the ASCII comma
+      val nameString = parts[2].replace(COMMA_URL_ENCODING, ",")
+
+      if (lat == null || long == null || nameString.isBlank()) {
+        throw IllegalArgumentException(PARSE_ERROR_MESSAGE)
+      }
+
+      return Location(lat, long, nameString)
+    }
```
francelu, line 64: Shouldn't this function be outside of the companion object ? Why do we even need a companion object in the data class of `Location`?

app/src/test/java/com/android/periodpals/model/location/LocationDataClassTest.kt
```diff
@@ -0,0 +1,125 @@
...
+    /* Regex to verify the format
+    ^     start of string
+    -?    optional negative sign (coordinate may be negative)
+    \d+   one or more digits
+    \.\d+ decimal point followed by one or more digits
+    ,     a literal comma
+    \S+   one or more non-whitespace characters
+    .*    allows for more characters in case the name has multiple words
+    $     end of string
+    */
```
francelu, line 35: A bit weird to have all this explanation in the middle of the test's code. I think it would be better to have it in the documentation of this test.

francelu (approved):
## Summary
The PR effectively consolidates the `GPSLocation` and `Location` data classes, addressing the cumbersome conversions and potential bugs caused by having two similar classes. Moreover, the addition of serialisation methods prepares the codebase for upcoming features like Alert VM integration. The draft issue for backwards search is a thoughtful inclusion.


## Important

### Code Quality
- Good clean-up of the `GPSLocation` class and replacing it with `Location`, improving maintainability and reducing redundancy.
- The addition of serialisation methods (`toString()` and `fromString()`) ensures that the `Location` class is ready for future use cases like Alert VM integration.

### Functionality
- The refactoring appears consistent and aligns with the objective of replacing `GPSLocation` without introducing regressions.
- The default handling for the `name` attribute when using the "Current location" option is clear and functional.

### Documentation
- The note explaining the temporary handling of "Current location" is helpful, and the draft issue for the backwards search method is a good forward-looking step.


## Testing
- The new tests for the serialisation methods look comprehensive and ensure correctness.


## Suggestions
- Address the code comments added during the review process to resolve potential questions.


## Steps Before PR Merged
- Resolve the code comments flagged for both @agonzalez-r and me.
- Good job overall ! Keep up the good work :)

app/src/test/java/com/android/periodpals/model/location/LocationDataClassTest.kt
agonzalez-r: Okay I see, in that case you try should using the function `.any()`, instead of the for loop, to check all the elements of the list.

app/src/main/java/com/android/periodpals/model/location/Location.kt
lazarinibruno: Good catch!

app/src/main/java/com/android/periodpals/model/location/Location.kt
lazarinibruno: Oki

app/src/main/java/com/android/periodpals/model/location/Location.kt
lazarinibruno: `fromString()` is a class function and not an instance function. The equivalent of defining static methods in Kotlin is by defining them inside the `companion object`.

app/src/test/java/com/android/periodpals/model/location/LocationDataClassTest.kt
lazarinibruno: I agree with you that it is unusual to have that big explanation there, but I added it since (as far as I know) none of us really knows how to work with regex expressions. Having it there explains what that line of code is doing, which would be otherwise obscure.

However, I've never seen a test with doc. I think is better to leave it there.

app/src/main/java/com/android/periodpals/model/location/Location.kt
francelu: Okay I see.

app/src/test/java/com/android/periodpals/model/location/LocationDataClassTest.kt
francelu: You decide :)  
I think it's possible to have documentation for tests, but only when it's a complex one. Usually the test name should be good enough to not need documentation which is your case here.

lazarinibruno: Okayyy I prefer leaving it as-is then.

PR #247, opened 2024-11-26
Feat/alert/create and edit alert vm integration: Integration of Alert VM to CreateAlert and EditAlert

agonzalez-r:
# Integration of Alert ViewModel to CreateAlert and EditAlert

## Description
**Important**: this task was originally divided in two (one for CreateAlert and one for EditAlert). However, I had to do modifications on `AlertComponents.kt` for the code to be more logical, therefore this changes reflect on **both** screens and it is not possible for the CI/tests to pass if not both are adapted. 
**TLDR:** Had to modify both screens (create and edit) in the same PR, modifications in both are _very_ similar. 

This PR introduces the final implementation for `CreateAlert` and `EditAlert` by adding their respective call to `AlertViewModel` and ensuring their actions are reflected in the Supabase database: 
- For `CreateAlert` it creates the alert in the table 
- For `EditAlert`, it updates the alert in the table if click on `Save` or deletes the alert if click on `Delete`

In `MainActivity.kt` I initialise an instance of `AlertViewModel` and pass it as parameter to `CreateAlertScreen()` (properly thanks to the refactoring of the view model).

I modified `AlertComponents.kt` for the code to be more consistent, I removed useless variables and return values from some functions. 
 For the Product field and Urgency field, I created parse functions to convert text to `Product` and respectively `Urgency`. This functions were added to `Alert.kt` and tested. 

I encountered troubles with the repository when using the previous version of the data class which left the parameters `id` and `createdAt` null, in order for it to be initialised in supabase. It turns out it doesn't work like that, if supabase receives the value as null, it will not generate the value itself. Therefore, the best solution I found was to generate the values directly in the data class. It seems to work fine, I verified all the fields in supabase are correctly filled. 

I also had to adapt `AlertDto` and `AlertModelSupabase` to the fact that `id` and `createdAt` can no longer be null.

Finally, I added the permission request pop up for the location (the same that is used in `Map.kt`), because the Location field option "Current Location" needs this permission. 

#### Added
None 

#### Modified
- `main/.../MainActivity.kt`
- `main/.../ui/alert/CreateAlert.kt`
- `main/.../ui/alert/EditAlert.kt`
- `main/.../ui/components/AlertComponents.kt`
-  `main/.../model/alert/Alert.kt`
- `main/.../model/alert/AlertDto.kt`
- `main/.../model/alert/AlertModelSupabase.kt`

#### Removed
None

## Dependencies Added
None

## Testing
Added test for the parse functions I added to `Alert.kt`
Modified `CreateAlertScreenTests` and `EditAlertScreenTests`, mainly their call to the screen functions (because new parameters were added to them, to pass the needed view models). 

**Important**: for `EditAlertScreenTests` I noticed that after adding the `alert.location` as the value for the Location field before updating it, it is impossible to left such field empty. If no new location is selected, then the previous location will remained unchanged. So the previous test `updateAlertInvalidLocation` that verified that it was not a valid input, was renamed `updateAlertEmptyLocation` and verifies that it does indeed accept to as valid input, as this is the expected behaviour. In another task we just need to make sure the user knows that the previous location was saved. 
All tests pass.


app/src/main/java/com/android/periodpals/model/alert/AlertDto.kt
taghizadlaura, line 1: LGTM !

app/src/main/java/com/android/periodpals/ui/components/AlertComponents.kt
taghizadlaura, line 1: Great job on the clean-up and refactoring!

app/src/test/java/com/android/periodpals/model/alert/AlertTests.kt
taghizadlaura, line 1: LGTM!!

As I mentioned in the global review, I'm not sure why, but the code coverage is going down. It's not a big deal, but it would be great if you could increase it before merging. I’m not exactly sure which tests to add, so if you don’t have any ideas either, it’s fine to leave it as is. Since there are still end-to-end tests, it should be okay.

taghizadlaura (approved):
## Summary
I'm sorry, but I feel like I don't fully understand all the changes since I haven't used Supabase before. I may not be the best person to review this code, but since it seems to work as expected and I didn't find anything strange, I approve it!

LGTM :
This PR seems to effectively integrates AlertViewModel with CreateAlert and EditAlert screens, ensuring that alert creation, updating, and deletion operations are properly reflected in the Supabase database. Significant refactoring was done in AlertComponents.kt for improved consistency. Parsing functions for converting text representations of Product and Urgency into their respective enums were added to Alert.kt. Additionally, AlertDto and AlertModelSupabase were updated to handle non-nullable id and createdAt fields, and a location permission request was added for the "Current Location" option.

## Important

### Code Quality
LGTM :
- **Readability & Consistency**: The code is clear, readable, and logically structured. The refactor in AlertComponents.kt is well-executed. The new parsing functions `textToProduct` and `textToUrgency` in Alert.kt are well-documented with detailed comments explaining their purpose.
- **Simplification**: The parsing functions are straightforward and well-integrated into the Alert.kt file.

### Functionality
LGTM :
- **Correctness**: The core functionalities for creating, updating, and deleting alerts seem to work as intended. 

### Performance 
LGTM : 
No performance issues.

### Security 
LGTM :
No security issues were observed.

## Testing
LGTM :
The parsing functions in Alert.kt are well-tested, along with the updated CreateAlertScreenTests and EditAlertScreenTests that validate the main functionalities.
I don't know exactly why, but the code coverage, even though it is still high, has dropped below 70%. It's not a big deal, but it would be nice if you could increase it.

## Documentation
LGTM :
Everything is well-documented and the PR description effectively explains the purpose of the changes and any relevant details.

## Suggestions
I don't have any specific suggestions, but the test coverage is going down. It would be great if you could add some tests.

## Steps before PR approved
LGTM please answer my comments !

lazarinibruno (approved):
## Summary

Thanks for your contribution! It is very satisfying to finally have the `CreateAlertScreen` and `EditAlertScreen` work with the backend 👾🛜.

## Important

As you mentioned, it would improve the UX that whenever the user modifies the `LocationField` but doesn't choose an option from the Dropdown Menu in the Edit Alert Screen, the Alert's location is shown again.

### Code Quality

Overall, code is readable and clear.

### Functionality

App works perfectly when running locally and tests also pass locally.

## Testing

- I think that it is important to review with the coaches whether we should test the uncovered lines in `CreateAlert.kt` and `EditAlert.kt`. Its true that tests are already done for the view models, but I don't know if the `onSuccess` and `onFailure` should be tested from the context of those screens.
  
- Besides that, the test remain basically the same except for updating the signatures of `CreateAlertScreen` and `EditAlertScreen`.
  

## Documentation

- You updated the doc accordingly and added pertinent doc for the new code!
 
- Your PR goes over the most important changes and explains your design choices, which is good. I think however that the structure and writing could be a tad better.
  
## Suggestions

My main suggestion is refactoring the `loadAuthenticationUserData` and `loadUser` in `CreateAlertScreen` to be inside the `LaunchedEffect(Unit)`.

Code within a composable is re-run every time the screen is recomposed (this happens, for instance, when the user types in a text field).

From my understanding, these are functions that should be executed once (when the screen composes for the first time). So putting them in the `LaunchedEffect` ensures that they are run only once and not on every recomposition.

## Steps before PR approved

As usual, very solid work! Please consider my nitpicky comments. Otherwise, looks good to merge :D

app/src/main/java/com/android/periodpals/ui/alert/CreateAlert.kt
```diff
@@ -51,23 +64,60 @@ private const val DEFAULT_MESSAGE = ""
...
+    authenticationViewModel: AuthenticationViewModel,
+    userViewModel: UserViewModel,
 ) {
   val context = LocalContext.current
-  var message by remember { mutableStateOf(DEFAULT_MESSAGE) }
+  var product by remember { mutableStateOf<Product?>(null) }
+  var urgency by remember { mutableStateOf<Urgency?>(null) }
   var selectedLocation by remember { mutableStateOf<Location?>(null) }
+  var message by remember { mutableStateOf(DEFAULT_MESSAGE) }
+  authenticationViewModel.loadAuthenticationUserData(
```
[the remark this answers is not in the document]
agonzalez-r: True! I didn't think about the `LaunchedEffect` before thinking about the permissions pop up, but it is also a good solution for loading the data

app/src/main/java/com/android/periodpals/ui/components/AlertComponents.kt
```diff
@@ -65,51 +65,33 @@ private const val LOCATION_FIELD_TAG = "AlertComponents: LocationField"
...
- * @return A boolean indicating whether a product is selected.
  */
 @Composable
-fun productField(
-    product: String,
-    onValueChange: (String) -> Unit,
-    isSelected: Boolean = false
-): Boolean {
-  val (productIsSelected, setProductIsSelected) = remember { mutableStateOf(isSelected) }
+fun ProductField(product: String, onValueChange: (String) -> Unit) {
```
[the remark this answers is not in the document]
agonzalez-r: actually I learned that with my last PR where I implemented `AlertComponents`, but when a function has a return value, standard is that it doesn't begin with a capital letter. Since I removed the return value, I can put a capital letter!

app/src/main/java/com/android/periodpals/model/alert/Alert.kt
```diff
@@ -49,6 +51,36 @@ enum class Status {
   SOLVED, // The alert has been resolved, help was provided
 }
 
+/**
+ * Converts a text representation of a product to a `Product` enum.
+ *
+ * @param text The text representation of the product.
+ * @return The corresponding `Product` enum, or `null` if the text does not match any product.
+ */
+fun textToProduct(text: String): Product? {
```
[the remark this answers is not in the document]
agonzalez-r: I think it's better not to use reserved names like "string"

app/src/test/java/com/android/periodpals/model/alert/AlertTests.kt
agonzalez-r: I discussed this in person with @lazarinibruno, most of the added lines of code where the use of functions of other classes (the view model, Alert or Location parsing, etc.). Therefore those functions are tested in their respective classes and _I don't think they should also be tested here_. 
I think the way to increase coverage for `CreateAlert`, `AlertLists` and `EditAlert` is with end2end tests. To be discussed for next sprint!

lazarinibruno:
> My main suggestion is refactoring the `loadAuthenticationUserData` and `loadUser` in `CreateAlertScreen` to be inside the `LaunchedEffect(Unit)`.
> 
> Code within a composable is re-run every time the screen is recomposed (this happens, for instance, when the user types in a text field).
> 
> From my understanding, these are functions that should be executed once (when the screen composes for the first time). So putting them in the `LaunchedEffect` ensures that they are run only once and not on every recomposition.

My bad, this introduced a bug! I did not think it thoroughly. Having a look into the `loadAuthenticationUserData` and `loadUser` would have showed me that there is already something similar being done with `viewModelScope.launch { ... }`. Apologies @agonzalez-r !!

Issue #249, opened 2024-11-27
Refactor `userID` in `AlertViewModel` to facilitate implementation


coaguila:
## Description
The objective is to refactor the `deleteAlert` function in `AlertViewModel` to facilitate the vm impementation. The will modify a couple of lines in a couple of files. The related tests will need to be updated as well.

## Done
Specific, measurable criteria to confirm the task is complete.
- Needs to able to be initialized without needing a userID
- Maintain consistent coding practices and quality
- Refactor tests affected by the incomming changes to maintain test coverage

## Steps to Completion
Break down the task into actionable steps. Make sure those steps are separable so that you can easily create subtasks/subissues from these:
- [ ] Remove `userId` parameter
- [ ] Add alternative to `userId`
- [ ] Fix incorrect tests

## Points to Consider / Known Risks
Optional: List any potential risks, technical constraints, or dependencies. Mention limitations or assumptions, such as “Requires access to external API” or “Limited by current design framework,” if applicable.

PR #250, opened 2024-11-27
Refactor `userId` parameter in `AlertViewModel`

coaguila:
# Refactor `userId` parameter in `AlertViewModel`

## Description
This PR refactors the `userId` private parameter into a new private state `userId` that can be set. It closes issue #249. This code should be revisited once a new `SharedContext` data class is shared. This is a simple patch fix. `userId` state's default value is null.

## Changes
Added new setter function `setUserID` for the private mutable state `userId` in  `AlertViewModel.kt` and the new private mutable state is not ideal but works and avoids any innitialization issues when implementing.
#### Modified
- `app/src/main/java/com/android/periodpals/model/alert/AlertViewModel.kt`
- `app/src/test/java/com/android/periodpals/model/alert/AlertViewModelTest.kt`

## Testing
Refactored how view model is innitialized.

### DIFF
--- a/app/src/main/java/com/android/periodpals/model/alert/AlertViewModel.kt
+++ b/app/src/main/java/com/android/periodpals/model/alert/AlertViewModel.kt
@@ -173,4 +173,13 @@ class AlertViewModel(
   fun setFilter(filter: (Alert) -> Boolean) {
     viewModelScope.launch { alertFilter.value = filter }
   }
+
+  /**
+   * Sets the `userID` state of View Model
+   *
+   * @param uid user id to be stored in a private state
+   */
+  fun setUserID(uid: String) {
+    viewModelScope.launch { userId.value = uid }
+  }
 }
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/android/periodpals/model/alert/AlertViewModel.kt
```diff
@@ -25,17 +25,19 @@ private const val TAG = "AlertViewModel"
  * @property _filterAlerts Mutable state holding the list of alerts filtered by `alertFilter`
  * @property filterAlerts Public state exposing the list of alerts filtered y `alertFilter`
  */
-class AlertViewModel(
-    private val alertModelSupabase: AlertModelSupabase,
-    private val userId: String
-) : ViewModel() {
+class AlertViewModel(private val alertModelSupabase: AlertModelSupabase) : ViewModel() {
```
agonzalez-r, line 28: Please update documentation: 
- remove `uid` parameter from documentation
- add a little comment about when/why to call `setUserId()`, for better understanding

agonzalez-r (approved):
LGTM! Thank you
As you explain in the PR description, this is not an ideal fix but functional, and that's what we need. It avoids the issues encountered with the Alert VM innit. 
However, I think it's not very intuitive to know when to use the `setUserId()` and why it's necessary to use it, so I do think it's important a little more explaining in the documentation, for easier and proper usage of this solution.

Harrish92 (approved):
Could you describe in more details what problem this PR fixes? Other than this code LGTM.

Issue #264, opened 2024-11-30
Integrate form validation library into profile screens


charliemangano:
## Description
This is a subtask of #261. The goal is to implement form validation via the FromBuilder library into the profile screens (CreateProfile and EditProfile)

## Done
- fields are properly validated using the library
- already implemented tests still pass

Issue #261, opened 2024-11-29
Integrate API for form validation


charliemangano:
### Description:
We want to have the form validation (sign in, signup, create profile, create alert) be done by an external API. It would standardize and make easier the form validation.

### Done:
- **Functional Requirements:** 
  - working integration of an external API for form validation
  - the existing functionality should stay the same but be managed by the API 
- **Quality Standards:**  
  - clean code
  - documentation
- **Testing Requirements:** 
  - tests already implemented should still pass
  - 80% coverage

### Preconditions:
Requires research to choose and integrate the external API.

### Steps to Completion:
- [x] research
- [ ] integration
- [ ] testing

charliemangano:
We will be using [FormBuilder](https://github.com/jkuatdsc/form-builder), which is a form validation library that integrates well with Jetpack Compose

PR #265, opened 2024-11-30
Feat/profile/form-validation : Integrate form validation library into profile screens

charliemangano:
# Integrate form validation library into profile screens

## Description
This PR integrate a new library [FromBuilder](https://github.com/jkuatdsc/form-builder) to do form validation in our profile screens. This library is made to work with Jetpack Compose. This tremendously cleans up our screens since the validation logic is handled elsewhere, and better separates the duties MVVM-wise.
This PR closes issue #264 and part of #261.

## Changes
**Explanation of the library:** see [in previous PR](https://github.com/PeriodPals/periodpals/pull/263)

**Actual changes:**
I set up the validation logic in the `UserViewModel`: the `FormState` is accessible as an attribute of the class and is associated with the corresponding `TextFieldState`s and `Validator`s.
In the screens, I replaced the existing states with the `formState` and offshoots `nameState`, `dobState` etc. I also replaced the validation logic with calls to `validate` and the `onChange` logic with calls to `change`. The error messages are also better handled using the `errorMessage` function.

Seeing the feedback from M2, I added a check for the length of the inputs.

#### Added
None
#### Modified
- `app/src/main/java/com/android/periodpals/model/user/UserViewModel.kt`: implemented the validation logic
- `app/src/main/java/com/android/periodpals/ui/profile/CreateProfile.kt`, `app/src/main/java/com/android/periodpals/ui/profile/EditProfile.kt` and `app/src/main/java/com/android/periodpals/ui/components/ProfileComponents.kt`: integrated the library into the screens
- `app/src/test/java/com/android/periodpals/model/user/UserViewModelTest.kt`, `app/src/test/java/com/android/periodpals/ui/profile/CreateProfileTest.kt` and `app/src/test/java/com/android/periodpals/ui/profile/EditProfileTest.kt`: update tests
- `app/build.gradle.kts`, `settings.gradle.kts` and `gradle/libs.versions.toml`: added FormBuilder dependency

#### Removed
None

## Dependencies Added
- `com.github.jkuatdsc:form-builder`: FormBuilder dependency

## Testing
I updated the tests to take into account this new library: in `UserViewModelTest`, I implemented new checks that verify the `FormState` attribute values, and in `CreateProfileTest` and `EditProfile`, I provided the `FormState` as part of the `UserViewModel` mock.

### DIFF
--- a/app/src/main/java/com/android/periodpals/model/user/UserViewModel.kt
+++ b/app/src/main/java/com/android/periodpals/model/user/UserViewModel.kt
@@ -1,23 +1,71 @@
 package com.android.periodpals.model.user
 
+import android.icu.util.GregorianCalendar
 import android.util.Log
 import androidx.compose.runtime.State
 import androidx.compose.runtime.mutableStateOf
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
+import com.dsc.form_builder.FormState
+import com.dsc.form_builder.TextFieldState
+import com.dsc.form_builder.Validators
 import kotlinx.coroutines.launch
+
+private const val TAG = "UserViewModel"
+
+private const val MAX_NAME_LENGTH = 128
+private const val MAX_DESCRIPTION_LENGTH = 512
+
+private const val ERROR_INVALID_NAME = "Please enter a name"
+private const val ERROR_NAME_TOO_LONG = "Name must be less than $MAX_NAME_LENGTH characters"
+private const val ERROR_INVALID_DESCRIPTION = "Please enter a description"
+private const val ERROR_DESCRIPTION_TOO_LONG =
+    "Description must be less than $MAX_DESCRIPTION_LENGTH characters"
+private const val ERROR_INVALID_DOB = "Invalid date"
+
+private val nameValidators =
+    listOf(
+        Validators.Required(message = ERROR_INVALID_NAME),
+        Validators.Max(message = ERROR_NAME_TOO_LONG, limit = MAX_NAME_LENGTH),
+    )
+private val descriptionValidators =
+    listOf(
+        Validators.Required(message = ERROR_INVALID_DESCRIPTION),
+        Validators.Max(message = ERROR_DESCRIPTION_TOO_LONG, limit = MAX_DESCRIPTION_LENGTH),
+    )
+private val dobValidators =
+    listOf(
+        Validators.Required(message = ERROR_INVALID_DOB),
+        Validators.Custom(message = ERROR_INVALID_DOB, function = { validateDate(it as String) }),
+    )
+private val profileImageValidators = emptyList<Validators>()
 
 /**
  * ViewModel for managing user data.
  *
  * @property userRepository The repository used for loading and saving user profiles.
  */
-private const val TAG = "UserViewModel"
-
 class UserViewModel(private val userRepository: UserRepositorySupabase) : ViewModel() {
+  companion object {
+    const val NAME_STATE_NAME = "name"
+    const val DESCRIPTION_STATE_NAME = "description"
+    const val DOB_STATE_NAME = "dob"
+    const val PROFILE_IMAGE_STATE_NAME = "profile_image"
+  }
 
   private val _user = mutableStateOf<User?>(null)
   val user: State<User?> = _user
+
+  val formState =
+      FormState(
+          fields =
+              listOf(
+                  TextFieldState(name = NAME_STATE_NAME, validators = nameValidators),
+                  TextFieldState(name = DESCRIPTION_STATE_NAME, validators = descriptionValidators),
+                  TextFieldState(name = DOB_STATE_NAME, validators = dobValidators),
+                  TextFieldState(
+                      name = PROFILE_IMAGE_STATE_NAME, validators = profileImageValidators),
+              ))
 
   /**
    * Loads the user profile and updates the user state.
@@ -30,7 +78,7 @@ class UserViewModel(private val userRepository: UserRepositorySupabase) : ViewMo
       onSuccess: () -> Unit = { Log.d(TAG, "loadUser success callback") },
       onFailure: (Exception) -> Unit = { e: Exception ->
         Log.d(TAG, "loadUser failure callback: $e")
-      }
+      },
   ) {
     viewModelScope.launch {
       userRepository.loadUserProfile(
@@ -60,7 +108,7 @@ class UserViewModel(private val userRepository: UserRepositorySupabase) : ViewMo
       onSuccess: () -> Unit = { Log.d(TAG, "saveUser success callback") },
       onFailure: (Exception) -> Unit = { e: Exception ->
         Log.d(TAG, "saveUser failure callback: $e")
-      }
+      },
   ) {
     viewModelScope.launch {
       userRepository.upsertUserProfile(
@@ -74,7 +122,8 @@ class UserViewModel(private val userRepository: UserRepositorySupabase) : ViewMo
             Log.d(TAG, "saveUser: fail to save user: ${e.message}")
             _user.value = null
             onFailure(e)
-          })
+          },
+      )
     }
   }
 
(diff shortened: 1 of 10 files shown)


app/src/test/java/com/android/periodpals/ui/profile/EditProfileTest.kt
taghizadlaura, line 1: LGTM !

app/src/main/java/com/android/periodpals/ui/profile/EditProfile.kt
taghizadlaura, line 1: LGTM !

app/src/test/java/com/android/periodpals/model/user/UserViewModelTest.kt
taghizadlaura, line 1: LGTM ! Overall, well-structured and thorough tests !

app/src/test/java/com/android/periodpals/ui/profile/CreateProfileTest.kt
taghizadlaura, line 1: LGTM !

app/src/main/java/com/android/periodpals/ui/components/ProfileComponents.kt
```diff
@@ -193,29 +193,35 @@ fun ProfileInputDescription(description: String, onValueChange: (String) -> Unit
...
+    navigationActions: NavigationActions,
 ) {
 
   Button(
       modifier = Modifier.wrapContentSize().testTag(ProfileScreens.SAVE_BUTTON),
       onClick = {
-        val errorMessage = validateFields(name, dob, description)
+        val errorMessage =
+            when {
+              !dobState.validate() -> dobState.errorMessage
```
agonzalez-r, line 220: Good implementation here! remove all unnecessary const values for the error messages that are no longer used

app/src/main/java/com/android/periodpals/model/user/UserViewModel.kt
```diff
@@ -1,24 +1,72 @@
...
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
+import com.dsc.form_builder.FormState
+import com.dsc.form_builder.TextFieldState
+import com.dsc.form_builder.Validators
 import kotlinx.coroutines.launch
 
+private const val TAG = "UserViewModel"
+
+private const val MAX_NAME_LENGTH = 128
```
agonzalez-r, line 16: how did you choose these lengths ?

app/src/main/java/com/android/periodpals/model/user/UserViewModel.kt
```diff
@@ -1,24 +1,72 @@
...
+    listOf(
+        Validators.Required(message = ERROR_INVALID_DESCRIPTION),
+        Validators.Max(message = ERROR_DESCRIPTION_TOO_LONG, limit = MAX_DESCRIPTION_LENGTH),
+    )
+private val dobValidators =
+    listOf(
+        Validators.Required(message = ERROR_INVALID_DOB),
+        Validators.Custom(message = ERROR_INVALID_DOB, function = { validateDate(it as String) }),
+    )
+private val profileImageValidators = emptyList<Validators>()
```
agonzalez-r, line 41: Is there really nothing to verify here? if this is supposed to be an URI, can we verify that it is?

app/src/main/java/com/android/periodpals/ui/profile/EditProfile.kt
```diff
@@ -74,22 +72,26 @@ fun EditProfileScreen(userViewModel: UserViewModel, navigationActions: Navigatio
...
-  var dob by remember { mutableStateOf(userState.value?.dob ?: "") }
-  var description by remember { mutableStateOf(userState.value?.description ?: "") }
-  var profileImageUri by remember {
-    mutableStateOf(userState.value?.imageUrl ?: DEFAULT_PROFILE_PICTURE)
-  }
+  val formState = remember { userViewModel.formState }
+  formState.reset()
+
+  val nameState = formState.getState<TextFieldState>(UserViewModel.NAME_STATE_NAME)
+  nameState.change(userState.value?.name ?: "")
```
agonzalez-r, line 82: Here if the userState.value?.name is null, it sets the nameState to an empty string and idem for all the other parameters, isn't there a better "default" value to put instead of the empty string?

app/src/main/java/com/android/periodpals/model/user/UserViewModel.kt
```diff
@@ -104,7 +153,30 @@ class UserViewModel(private val userRepository: UserRepositorySupabase) : ViewMo
...
+  }
+}
+
+/**
+ * Validates the date is in the format DD/MM/YYYY and is a valid date.
+ *
+ * @param date The date string to validate.
+ * @return True if the date is valid, otherwise false.
+ */
+fun validateDate(date: String): Boolean {
```
agonzalez-r, line 168: good thinking having this function in the view model, I noticed that you didn't add any tests for it in `UserViewModelTests`, is it already tested somewhere else?

agonzalez-r (approved):
(I'm really sorry for the late review!)
## Review
I left a few questions and nitpicks in the comments, but overall: 
- Code is clean and consistent and properly documented. 
- Tests are correct, I just had a question about adding one  for `validateDate()` to `UserViewModelTests` (to ensure proper functionality of the function), but I'm not sure if it's already tested elsewhere. 

## Important remark / question 
The form validators you added throw an error when trying to save an input of a text field that doesn't comply with the length limits, however shouldn't we just not *let* the user input such long texts in the input field? 

**Example:**
Screenshot 2024-12-02 at 21 49 23

As you can see there is no limit as how long the input in the `name` text field is, here it is so long that the save button is no longer on the screen. I'm not sure if this was the scope of you task, but from what I understood, this is the problem mentioned in the feedback, we should still try to solve it.

app/src/main/java/com/android/periodpals/model/user/UserViewModel.kt
charliemangano: I seemed to be quite standard, and powers of 2 are nice :3 No real reason otherwise. If you have a better source for it I'll gladly change it but I couldn't find a length that everyone agrees on.

app/src/main/java/com/android/periodpals/model/user/UserViewModel.kt
charliemangano: The way we are doing images is gonna change since we're revisiting it. The form validation on it should be done by the person implementing the new way.

app/src/main/java/com/android/periodpals/ui/profile/EditProfile.kt
charliemangano: I did not choose this, it was also this way in the previous version. I believe there is already a Toast that appears. Plus in my opinion, having "Error loading your data" (or something similar) in the field makes the user have to delete all of it before inputting their changes, or else save "Error ..." as their name/description/..., which does not seem ideal.

charliemangano:
Hi @agonzalez-r, I'll answer your question here!
> The form validators you added throw an error when trying to save an input of a text field that doesn't comply with the length limits, however shouldn't we just not let the user input such long texts in the input field?
> Screenshot 2024-12-02 at 21 49 23
> As you can see there is no limit as how long the input in the name text field is, here it is so long that the save button is no longer on the screen.

That would indeed be a good addition to our app and would improve UX greatly! However since it's kind of outside of this scope and not necessary at this point, I propose we make an issue or a task about it instead of directly implementing it.

> I'm not sure if this was the scope of you task, but from what I understood, this is the problem mentioned in the feedback, we should still try to solve it.

I think as long as we keep the user from entering an infinite-length input, we're good with the feedback. However, as said, having it be immediate and interactive (directly upon user input as opposed to when clicking on "Save") would be great for UX and I really think we should try and implement it.

app/src/main/java/com/android/periodpals/model/user/UserViewModel.kt
agonzalez-r: Okay, we just need to be careful and don't forget it

app/src/main/java/com/android/periodpals/model/user/UserViewModel.kt
agonzalez-r: Thank you for completing the tests!

Issue #203, opened 2024-11-15
Improve map


lazarinibruno:
## Description
The GPS location works as well as the map! However, `osmdroid` is not the best library to use for different reasons. Among them, it is no longer maintained and its UI is quite old.

Besides that, the app does not prompt the user with some info if they decide to reject the location. We need to be able to tell the users that for the app to function, it needs access to their location

* [x] #115
* [x] #130
* [x] Only center the map in the user's location at the beginning
* [x] Add button to center map on the user
* [x] Implement light and dark mode
* [x] Use local tile source if map fails to request from API
* [x] Increase markers size
* [x] Add zoom in / out buttons

PR #289, opened 2024-12-07
Map improvements

lazarinibruno:
# Map improvements

## Description
This PR introduces a few minor yet relevant changes to the app. It closes issue #203.

Before the PO @francelu yells at me for doing my P2 task instead of the P1, I had to wait for @agonzalez-r to finish her [PR](https://github.com/PeriodPals/periodpals/pull/290) with the shared dialog composable to filter the locations. 

## Changes
- Map is able to change between dark mode and light mode
- ~If the app does not have internet connection, the map uses a local tile source~
- **Update**: OSMdroid doesn't have anything like a "local tile source", it always downloads the tiles from the internet. So the `NetworkChangeListener` is useless for the map, but maybe we can find another use for it.
- Increased the size of the markers and added a black outline (it was hard to distinguish them when they overlap)
- Added back the zoom in/out buttons (they show up when interacting with the map and then fade out)

To implement the second point, I added a `NetworkChangeListener` class which, as the name suggests, listens to the state of the network. In the spirit of the service classes that we have implemented, it exposes the state of the network through a StateFlow `isNetworkAvailable`. Therefore, it suffices to listen to it in screens that need to react to the state of the network. 

### Added
- `app/src/main/java/com/android/periodpals/services/NetworkChangeListener.kt`
- `app/src/test/java/com/android/periodpals/services/NetworkChangeListenerTest.kt`

### Modified
See files section.

### Removed
None.

## Dependencies Added
None.

## Testing
I wrote unit tests for the `NetworkChangeListener` class.

The coverage apparent coverage is low, but this is due to the fact that the MainActivity is not yet tested (will be done in the E2E tests):

image


app/src/main/java/com/android/periodpals/model/location/Location.kt
francelu, line 1: Good, it's more centred on campus now where our main typical user will be.

app/src/main/java/com/android/periodpals/ui/map/Map.kt
```diff
@@ -53,10 +55,10 @@ private const val MAX_ZOOM_LEVEL = 18.0
 private const val INITIAL_ZOOM_LEVEL = 17.0
 
 private const val CUSTOM_THEME_NAME = "Custom theme"
```
francelu, line 57: This is unused. Remove it ?

app/src/main/java/com/android/periodpals/ui/map/Map.kt
```diff
@@ -327,7 +328,12 @@ private fun setupCustomTileSource(
           return constructedUrl
         }
       }
-  mapView.setTileSource(customTileSource)
+
+  if (isNetworkAvailable) {
+    mapView.setTileSource(customTileSource)
+    return
+  }
+  mapView.setTileSource(TileSourceFactory.MAPNIK)
```
francelu, line 336: **Nitpick**: code understandability and cleanness ?
```suggestion
return if (isNetworkAvailable) mapView.setTileSource(customTileSource)
  else mapView.setTileSource(TileSourceFactory.MAPNIK)
```

app/src/main/java/com/android/periodpals/services/NetworkChangeListener.kt
francelu, line 1: You might want to add some documentation if you don't want to be beaten by the coaches :')

app/src/test/java/com/android/periodpals/services/NetworkChangeListenerTest.kt
francelu, line 1: LGTM. Tests pass !

francelu (changes requested):
## Summary
Good implementation.  

However, when I tried on emulator, I was on map (worked well), turned on air-plane mode (still kinda worked), went on another screen, came back to map, had ugly grid.
image

## Important
#### Code Quality
- Generally consistent and readable.

#### Functionality
- Doesn't work as intended in offline mode.

## Testing
As I mentioned, this big bug slipped through the net.

## Documentation
- A few missing...

## Suggestions
- Check comments.
- Fix offline map bug.

## Steps before PR approved
- Check comments.
- Fix offline map bug.

app/src/main/java/com/android/periodpals/ui/map/Map.kt
lazarinibruno: Good catch!

app/src/main/java/com/android/periodpals/ui/map/Map.kt
lazarinibruno: Yes, I wanted to avoid using the `else` statement but, in retrospective, you're right that using an `else` makes it clearer.

app/src/main/java/com/android/periodpals/services/NetworkChangeListener.kt
lazarinibruno: Hahaha, totally forgot about it. Thx.

coaguila (approved):
## Summary
As you said, `NetworkChangeListner` doesnt seem to be useful for this pr but is a welcomed addition because it does seem useful to have already even if there is no offline caching for maps. As you mentioned, this would create storage issues as well as osm not offering a solution, but if you are interested in storing locally, don't hesitate to look into the Room Jetpack Android Library.
## Important (which includes)
#### Code Quality
lgtm
#### Functionality
lgtm
## Testing
lgtm
## Documentation
Simple, quick, effecient
## Suggestions
Finish implementing France's requested changes.

app/src/main/java/com/android/periodpals/services/NetworkChangeListener.kt
francelu: Why do we still need `NetworkChangeListener` ? I don't think it's very meaningful to keep it in this PR even though we might need it for offline mode. TBD with @coaguila.

Same for the test class.

app/src/main/res/drawable/alert_marker.xml
francelu, line 1: Might want to adapt it in Sprint 10 to the different screen sizes using the `Type.kt` if we have some time left.

francelu (approved):
## Summary
LGTM.

## Important
#### Functionality
- Works seamlessly as intended.

## Steps before PR approved
- Look at last comments and resolve as needed.

app/src/main/java/com/android/periodpals/services/NetworkChangeListener.kt
lazarinibruno: Had I known that the OSMDroid does not support "local" tile sources, I obviously wouldn't have implemented it. But given that I already did, lets keep it. Otherwise, doing a separate PR for `NetworkChangeListener` would mean someone else reviewing code that you and and @coaguila already did.

app/src/main/res/drawable/alert_marker.xml
lazarinibruno: This is a good idea! However, it is not as straight-forward as it would seem. So lets leave it as a future task, as you suggest.

Issue #316, opened 2024-12-14
Secure location handling


charliemangano:
### Description
As of now, the location of each user is pushed to the `public.users` table, which is readable by anyone. This is a security concern, so we will change the way it works: we want to create a new table reserved for the users' location. This table will be writable to but not readable, so a user/dev cannot fetch this location. It should however be readable by our database function to be able to filter users and alert by their location.

### Done
- a secure way store the users' locations 
- locations are still properly uploaded
- locations are accessible to the database functions
- other features still work as expected

PR #350, opened 2024-12-19
Feat/geolocation/secure-location-handling : Secure location handling

charliemangano:
# Secure location handling

## Description
This PR introduces changes to the way we're handling the users' location so that it is not shared client-side. This ensures more privacy for our users. It closes issue #316.

## Changes
I created a new table on Supabase named `public.locations`, with RLS such that users can only create or update their location, but never read anything from the table. However, the database function that handles the users' locations for the alerts is given permission.

I implemented a model and view model for this new table. The model only has `create` and `update` functions. Using `upsert` here was not an option because it would have required us to give read permissions, which we didn't want to. To go with that, I also created a new data class `UserLocationDto` to be able to send data in the proper format to Supabase.

The lack of `upsert` is also the reason for the view model being a bit different from the other view models we have in the codebase: it has a single method `uploadUserLocation`. This function recreates the `upsert` behaviour without needing the same permissions: it first tries to create a new location in the table, but if that fails, it tries to update the possibly already present location.

Places that formerly used the user view model to upload the location were modified to use the new `UserLocationViewModel`.

#### Added
- `app/src/main/java/com/android/periodpals/model/location/UserLocationDto.kt`: the data transfert object for the user location
- `app/src/main/java/com/android/periodpals/model/location/UserLocationModel.kt` and `app/src/main/java/com/android/periodpals/model/location/UserLocationModelSupabase.kt`: interface and implementation of the model for the user location
- `app/src/main/java/com/android/periodpals/model/location/UserLocationViewModel.kt`: view model for the user location
- `app/src/test/java/com/android/periodpals/model/location/UserLocationDtoTest.kt`, `app/src/test/java/com/android/periodpals/model/location/UserLocationModelSupabaseTest.kt` and `app/src/test/java/com/android/periodpals/model/location/UserLocationViewModelTest.kt`: tests for aforementionned files
#### Modified
- `app/src/main/java/com/android/periodpals/services/GPSServiceImpl.kt`: adapted to call the new user location view model
- `app/src/main/java/com/android/periodpals/MainActivity.kt`: initialized user location logic properly and adapted call to `GPSServiceImpl`
- `app/src/main/java/com/android/periodpals/model/user/User.kt`, `app/src/main/java/com/android/periodpals/model/user/UserDto.kt` and `app/src/main/java/com/android/periodpals/model/user/UserModelSupabase.kt`: removed location attribute
- `app/src/test/java/com/android/periodpals/model/user/UserDtoTest.kt`, `app/src/test/java/com/android/periodpals/model/user/UserModelSupabaseTest.kt`, `app/src/test/java/com/android/periodpals/model/user/UserViewModelTest.kt` and `app/src/test/java/com/android/periodpals/services/GPSServiceImplTest.kt`: adapted existing tests
- `supabase/functions/push_new_alert/index.ts`: added logs
#### Removed
None
## Dependencies Added
None

## Testing
Implemented new tests and adapted old ones.

### DIFF
--- a/app/src/androidTest/java/com/android/periodpals/services/GPSServiceImplInstrumentedTest.kt
+++ b/app/src/androidTest/java/com/android/periodpals/services/GPSServiceImplInstrumentedTest.kt
@@ -9,8 +9,8 @@ import androidx.test.ext.junit.runners.AndroidJUnit4
 import androidx.test.rule.GrantPermissionRule
 import com.android.periodpals.model.authentication.AuthenticationViewModel
 import com.android.periodpals.model.location.Location
+import com.android.periodpals.model.location.UserLocationViewModel
 import com.android.periodpals.model.user.AuthenticationUserData
-import com.android.periodpals.model.user.UserViewModel
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.test.runTest
 import org.junit.After
@@ -36,7 +36,7 @@ class GPSServiceImplInstrumentedTest {
   private lateinit var activity: ComponentActivity
   private lateinit var gpsService: GPSServiceImpl
   private lateinit var authenticationViewModel: AuthenticationViewModel
-  private lateinit var userViewModel: UserViewModel
+  private lateinit var userLocationViewModel: UserLocationViewModel
 
   // Default location
   private val defaultLat = Location.DEFAULT_LOCATION.latitude
@@ -45,7 +45,7 @@ class GPSServiceImplInstrumentedTest {
   @Before
   fun setup() {
     authenticationViewModel = mock(AuthenticationViewModel::class.java)
-    userViewModel = mock(UserViewModel::class.java)
+    userLocationViewModel = mock(UserLocationViewModel::class.java)
 
     scenario = ActivityScenario.launch(ComponentActivity::class.java)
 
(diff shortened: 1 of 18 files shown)


coaguila (approved):
## Summary
Great new abstraction to change how we track users locations.
## Important (which includes)
#### Code Quality
All new code has kept the same standard of readability in the code base. The new supabase class is consistent in how the other supabase classes are coded.
#### Functionality
lgtm
#### Security (if applicable)
Making the table RLS to read only is great security practice.
## Testing
Adapts affected tests and adds relevant tests.
## Documentation
lgtm

lazarinibruno (approved):
# Summary

Thanks for the contribution, its a massive step in terms of our values for the app.

## Notes

- This implementation follows the Bell-LaPadula No Read-up security property - I think Carmela Troncoso would be proud of you 🥹🥹.
  
- Very clever workaround for the `upsert` function! It mimics it behaviour without needing to allow for read access.
  

## Code quality

- Consistent and well-written doc.
  
- Well written and clear code.
  

## Tests

- Good adaptation of already existing tests.
  
- Pertinent test additons.
  
- I played around for a while with the settings to make sure that they work.
  
- Tests pass locally.
  

## Steps before merge

LGTM.

